Issue #57, opened 2018-08-21
Create Super-Admin User


matthewschoen25:
**Admin**

As an admin, I can change an existing user's role from an admin to a super admin or vice versa in the view all users overview

- [ ] Current Admin can click on "Role" line item and use the drop down to change them to an admin or a super admin
- [ ] Ability for super and admin user changes to be made outside the database - front end for change requests   - TBD - best practice TBD
- [ ] Set up database

PR #66, opened 2018-08-29
Admin setup

tomlouismurphy:
### Ticket Number:

#4, #57 

Functionality has been added to allow administrator accounts to be viewed, created, to view information on the RLC site's users, and to create new admin accounts.

Landing page has been changed so that in addition to log in/registration options for users, there is now also an admin portal.

screen shot 2018-08-29 at 11 27 34 am

Viewing that admin portal only gives the option to log on as an admin, not to register a new account. The plan going forward is that initial admin accounts will be added through SQL queries, and then those admin accounts will have the responsibility of adding any further admin accounts to the site.

screen shot 2018-08-29 at 11 28 23 am

Upon sign in, an admin is taken to a new home screen. This page includes a link to screens where the admin account can update and delete itself (this functionality will be added in a future ticket), as well as expanded functionality for the Admin Portal, which now shows a list of all the website's users. In a future ticket, admin accounts will also be granted access to update and delete user accounts.

screen shot 2018-08-29 at 11 29 16 am

By clicking on "Create New Admin Account" on the admin home page, any admin with super-admin privileges can add a new admin account to the site, which could then be bestowed on whatever person the previous admin wants to have control of the new account. This feature is only given to those with super-admin status (controlled by a boolean value in the admins table) so an ordinary admin account would not even be able to view the "Create New Admin Account" link. So the below page should only be readable by super-admin accounts - any other user who attempt to access it will receive an error message.

screen shot 2018-08-29 at 11 30 42 am

Upon registering a new admin, a confirmation message will be given, but the super-admin will remain logged into their account rather than being logged into the new admin account they have just created.

screen shot 2018-08-29 at 11 32 27 am

### Checklist For Submitter

### Checklist For Reviewer


app/controllers/registrations_controller.rb
```diff
@@ -0,0 +1,18 @@
+class RegistrationsController < Devise::RegistrationsController
+
+  skip_before_action :require_no_authentication, only: [:new, :create]
```
EvanGrillo: From Tom in slack: "this line of code is necessary to allow admins to create an additional admin account while still logged in as admins. If this code isn't there, it'll throw an error if I try to access the admins/sign_up route while currently logged in"

app/models/admin.rb
```diff
@@ -0,0 +1,7 @@
+class Admin < ApplicationRecord
+  # Include default devise modules. Others available are:
+  # :confirmable, :lockable, :timeoutable and :omniauthable
+  devise :database_authenticatable, :registerable,
+         :recoverable, :rememberable, :trackable, :validatable
+  validates	:email, uniqueness: true
```
EvanGrillo: A note that uniqueness only applies to admin and that an address can also be used for a user account. When brought up, this led to a discussion about why we have an admin and user model. We agreed to bring this up for discussion with Jason (Technical Lead)

app/views/admins/registrations/new.html.erb
```diff
@@ -0,0 +1,49 @@
...
+
+      <div class="field">
+        <%= f.label :email %><br />
+        <%= f.email_field :email, autofocus: true, autocomplete: "email" %>
+      </div>
+
+      <div class="field">
+        <%= f.label :password %>
+        <% if @minimum_password_length %>
+        <em>(<%= @minimum_password_length %> characters minimum)</em>
```
EvanGrillo: this password length check could be handled in the admin model file

spec/features/admins_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'rails_helper'
+require 'spec_helper'
+require 'faker'
```
EvanGrillo: noting in case gem faker is taken out.

spec/features/admins_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'rails_helper'
```
EvanGrillo: this package looks like it could potentially be vulnerable and left unmaintained 
screen shot 2018-09-05 at 8 28 42 pm

spec/models/admin_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+require 'spec_helper'
+require 'faker'
+
+RSpec.describe Admin, :type => :model do
```
EvanGrillo: wondering if this is needed now that we have a method to create super admins remotely if needed

Gemfile
```diff
@@ -21,6 +21,9 @@ gem 'devise'
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', '>= 1.1.0', require: false
 
+# library for rspec testing
+gem 'faker'
```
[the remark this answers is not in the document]
tomlouismurphy: I don't think the gem is potentially dangerous at this time, and it's only employed in putting together automated testing. I do seewhat you mean though in terms of ensuring that tests are still valid as we add more things to user/admin schema and controllers, and I'll keep an eye on that and likely rerun/rewrite tests as we move forward.

app/controllers/registrations_controller.rb
tomlouismurphy: Added a comment in the code to explain this

app/views/admins/registrations/new.html.erb
tomlouismurphy: at present it looks like the password length check is handled the same way with the user views and model, so I'd favor leaving this as is and staying consistent. but I can also see how handling it in the model would give certain advantages, ie, wouldn't have to repeat code when writing a view to change passwords. won't change anything now but open to doing so on further conversation

spec/features/admins_spec.rb
tomlouismurphy: rails_helper isn't a gem - this is just requiring the rails_helper.rb file located in the spec folder that contains all our tests

spec/models/admin_spec.rb
tomlouismurphy: it's true that the fact that admin registration is now accomplished through seeding the database rather than registration on the site is significant, but I think these tests are more important in ensuring the admin schema is sufficient, and I'd therefore argue they're important and should stay

Issue #73, opened 2018-10-04
Migration error 


matthewschoen25:
the first file creates columns such as confirmation_token
then second file also creates the token. (same columns) 

***delete duplicate columns

one needs to be deleted 
(throughout the scripts makes sure that the columns are only created once.)

bottleneck created because of script ordering.  

add a CI script to check migration order

EvanGrillo:
I think the cleanest move would be to destroy migration ```20180820173442_add_confirmable_to_devise.rb```

Therefore, when creating the database & User model, the Devise confirmable module will be included and not need to be enabled in a later migration. It should also be noted that we don't have a local environment variable method yet nor a domain/address/password for the mailer in development.

My memory is hazy but I think one of two things happened:

1. I ran the ```20180807014553_devise_create_users.rb``` migration without the confirmable module & then edited the file after to include it. The edited migration file was included in the pull request.  

2. I ran that migration with the confirmable module, created and ran a new migration to remove it to login into an account without confirmation, and then destroyed that migration.

PR #75, opened 2018-10-04
Fixed config and addressed docker issues

zachgarwood:
This PR fixes some conflicting configurations / migrations we had, as well as some of the finicky bits of the docker setup.


zachgarwood:
This closes #73

PR #79, opened 2018-10-11
Addressing Webpacker / React

zachgarwood:
These changes address getting Webpacker set up for React. The short version is that webpacker looks for "packs" of code `/frontend/packs`. For this project, we're probably only ever going to have one pack called "application". The application pack will simply be a registry for all of the components that we create. Components are made up of a template (.erb), styling (.css), and functionality (.js) in `frontend/components/<component_name>/`.


zachgarwood:
Due to a slight change in direction, I'm going to close this PR.

PR #127, opened 2019-01-23
Recipe seeds update

mareann:
Create Recipe and Tag tables model including seed rows
Add awesome_print gem for rails console

[rlc_tables.docx](https://github.com/the-difference-engine/rescue-leftover-cuisine2/files/2785755/rlc_tables.docx)

Files added or modified
#model files
https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/RecipeSeedsUpdate/app/models/recipe.rb
https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/RecipeSeedsUpdate/app/models/tag.rb
#db files
https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/RecipeSeedsUpdate/db/seeds.rb
#db migrate files
https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/RecipeSeedsUpdate/db/migrate/20190121210000_create_recipes.rb
https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/RecipeSeedsUpdate/db/migrate/20190122160000_create_tags.rb

https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/RecipeSeedsUpdate/Gemfile
https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/RecipeSeedsUpdate/Gemfile.lock

### Checklist For Submitter

* [x ] I have documented all new functionality
* [ x] I have screenshoted new UI changes and attached them to this PR

### Checklist For Reviewer

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,10 @@ group :development do
   gem 'spring-watcher-listen', '~> 2.0.0'
 end
 
+group :development do
+  # awesome print for rails console
+  gem 'awesome_print'
+end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -43,6 +43,7 @@ GEM
       minitest (~> 5.1)
       tzinfo (~> 1.1)
     arel (9.0.0)
+    awesome_print (1.8.0)
     bcrypt (3.1.12)
     bootsnap (1.3.2)
       msgpack (~> 1.0)
@@ -169,6 +170,7 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
+  awesome_print
   bootsnap (>= 1.1.0)
   byebug
   devise (~> 4.5)
--- a/db/migrate/20190121210000_create_recipes.rb
+++ b/db/migrate/20190121210000_create_recipes.rb
@@ -1,0 +1,17 @@
+# Create Recipes Table
+
+class CreateRecipes < ActiveRecord::Migration[5.2]
+  def change
+    create_table :recipes do |t|
+      ## 
+      t.string :title,       null: false, default: ""
+      t.string :description, null: false, default: ""
+      t.text   :ingredients, array: true, default: []
+      t.text   :directions,  array: true, default: []
+      t.string :difficulty,  null: false, default: "Intermediate"
+      t.integer :duration,   null: false, default: 0
+      t.integer :servings,   null: false, default: 2
+      t.string  :photos,     array: true, default: []    
+    end
+  end
+end
--- a/db/migrate/20190121210000_create_recipes.rb
+++ b/db/migrate/20190121210000_create_recipes.rb
@@ -5,7 +5,8 @@ class CreateRecipes < ActiveRecord::Migration[5.2]
     create_table :recipes do |t|
       ## 
       t.string :title,       null: false, default: ""
-      t.string :description, null: false, default: ""
+      t.string :snipped,     null: false, default: ""
+      t.string :meal,        null: false, default: ""
       t.text   :ingredients, array: true, default: []
       t.text   :directions,  array: true, default: []
       t.string :difficulty,  null: false, default: "Intermediate"
--- a/db/migrate/20190122160000_create_tags.rb
+++ b/db/migrate/20190122160000_create_tags.rb
@@ -2,7 +2,7 @@
 
 class CreateTags < ActiveRecord::Migration[5.2]
   def change
-    create_table :recipes do |t|
+    create_table :tags do |t|
       ## 
       t.string :title,null: false   
     end
--- a/db/migrate/20190122160000_create_tags.rb
+++ b/db/migrate/20190122160000_create_tags.rb
@@ -1,0 +1,10 @@
+# Create Tags Table
+
+class CreateTags < ActiveRecord::Migration[5.2]
+  def change
+    create_table :recipes do |t|
+      ## 
+      t.string :title,null: false   
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_01_13_005107) do
+ActiveRecord::Schema.define(version: 2019_01_21_210000) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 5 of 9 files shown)


Gemfile
```diff
@@ -42,6 +42,10 @@ group :development do
   gem 'spring-watcher-listen', '~> 2.0.0'
 end
 
+group :development do
+  # awesome print for rails console
+  gem 'awesome_print'
```
vhoof: I think you could put this in the `group :development` block right above this one.

db/migrate/20190121210000_create_recipes.rb
```diff
@@ -0,0 +1,18 @@
...
+      ## 
+      t.string :title,       null: false, default: ""
+      t.string :snipped,     null: false, default: ""
+      t.string :meal,        null: false, default: ""
+      t.text   :ingredients, array: true, default: []
+      t.text   :directions,  array: true, default: []
+      t.string :difficulty,  null: false, default: "Intermediate"
+      t.integer :duration,   null: false, default: 0
+      t.integer :servings,   null: false, default: 2
+      t.string  :photos,     array: true, default: []
```
vhoof: I think it'd be good to add `t.timestamps` at the end here, which will add created_at, updated_at, and deleted_at timestamps.

db/migrate/20190121210000_create_recipes.rb
```diff
@@ -0,0 +1,18 @@
+# Create Recipes Table
+
+class CreateRecipes < ActiveRecord::Migration[5.2]
+  def change
+    create_table :recipes do |t|
+      ## 
+      t.string :title,       null: false, default: ""
+      t.string :snipped,     null: false, default: ""
```
vhoof: I think this should be `snippet` or `description`. I also think it's reasonable to let it be null and leave off the default. 

I also think removing the default on the title line would be good, too. We don't want that to be null, but we don't want blank titles, either.

db/migrate/20190121210000_create_recipes.rb
vhoof: I think it'd be good to leave off most/all of these defaults. Might be better to control that stuff in software, vs. at the database level. I also think most of these fields could be null (at least as far as the database is concerned), and we can handle validation further up the stack.

db/migrate/20190122160000_create_tags.rb
```diff
@@ -0,0 +1,10 @@
+# Create Tags Table
+
+class CreateTags < ActiveRecord::Migration[5.2]
+  def change
+    create_table :tags do |t|
+      ## 
+      t.string :title,null: false
```
vhoof: Can we put a space after the comma here to improve readability?

db/schema.rb
```diff
@@ -20,6 +20,17 @@
...
+    t.string "title", default: "", null: false
+    t.string "description", default: "", null: false
+    t.text "ingredients", default: [], array: true
+    t.text "directions", default: [], array: true
+    t.string "difficulty", default: "Intermediate", null: false
+    t.integer "duration", default: 0, null: false
+    t.integer "servings", default: 2, null: false
+    t.string "photos", default: [], array: true
+  end
+
```
vhoof: Looks like you need another `db:migrate` to do tags.

I think we'll also want to be able to join recipes and tags, so we'll need a join table and a many-to-many relationship. Might be worth checking out the intransitive association part of [this doc](https://www.sitepoint.com/master-many-to-many-associations-with-activerecord/).

db/seeds.rb
```diff
@@ -5,3 +5,18 @@
...
+
+#Comment.create(name: "Hal", text: "This is pretty good i guess", recipe_id: 1)
+
+# examples creating User row using rails console
+#User.create(email: "bob@domain.com", password: "Bob@RLC123", password_confirmation: "Bob@RLC123", username: "Bob@RLC", first_name: "Bob", last_name: "Domain", created_at: "2018-04-27 09:34:41.306138", updated_at: "2018-04-27 09:34:41.306138")
+#User.create(email: "joe@domain.com", password: "Joe@RLC123", password_confirmation: "Joe@RLC123", username: "Joe@RLC",first_name: "Bob", last_name: "Domain", created_at: "2019-01-20 09:34:41.306138", updated_at: "2018-01-20 09:34:41.306138")
+
+# examples creating Tag row using rails console
+#Tag.create(title: "Easy")
+#Tag.create(title: "Intermediate")
```
vhoof: We should probably create a couple seeds here and remove the commented stuff otherwise.

Gemfile
mareann: moved awesome_print above

db/migrate/20190122160000_create_tags.rb
mareann: inserted space

Gemfile.lock
```diff
@@ -187,4 +189,4 @@ RUBY VERSION
    ruby 2.5.3p105
 
 BUNDLED WITH
-   1.17.3
+   2.0.1
```
vhoof: I think we want to avoid bundler 2 for now, as it's backwards incompatible. So I think we may need to regenerate this file.

```
gem uninstall bundler      # pick the 2.0.1 one if you have options
gem install bundler -v 1.17.3
rm Gemfile.lock
bundle install
```

Alternatively, you could do this in Docker, with `docker-compose run web bundle install`.

db/migrate/20190123130000_recipe_has_many_tags.rb
```diff
@@ -0,0 +1,8 @@
+# Create Join Recipe Tags Table
+
+class Recipe < ActiveRecord::Migration[5.2]
+  def change
+     has_many :tags
```
vhoof: Hmm. I think this might work, but it might be better to use the `has_and_belongs_to_many` association instead, as described in https://guides.rubyonrails.org/association_basics.html#the-has-and-belongs-to-many-association and https://guides.rubyonrails.org/association_basics.html#creating-join-tables-for-has-and-belongs-to-many-associations. In particular, I like the separate indexes on the join table, since you'll rarely be searching for them together.

db/seeds.rb
```diff
@@ -5,3 +5,16 @@
...
 #   Character.create(name: 'Luke', movie: movies.first)
+
+# examples creating Recipe row using rails console
+Recipe.create(title: "Lasagna", snippet: "Lots of tomato sauce and cheese and pasta", ingredients: ["eggs","pasta","cheese"], directions: ["put ingredients in bowl","stir","put in pan","bake it"], difficulty: "Intermediate",duration:30,servings:4)
+Recipe.create(title: "Cookies", snippet: "chocolate chip heaven", ingredients: ["flour","eggs","chocolate chips"], directions: ["stir ingredients","make 1 inch balls","place on cookie sheet 2 inches apart","bake at 350 degress"], difficulty: "Beginner",duration:45,servings:6)
+
+# examples creating User row using rails console
+User.create(email: "bob@domain.com", password: "Bob@RLC123", password_confirmation: "Bob@RLC123", username: "Bob@RLC", first_name: "Bob", last_name: "Domain", created_at: "2018-04-27 09:34:41.306138", updated_at: "2018-04-27 09:34:41.306138")
+User.create(email: "joe@domain.com", password: "Joe@RLC123", password_confirmation: "Joe@RLC123", username: "Joe@RLC",first_name: "Bob", last_name: "Domain", created_at: "2019-01-20 09:34:41.306138", updated_at: "2018-01-20 09:34:41.306138")
```
vhoof: Based on the design docs, I think we want to use `easy`, `medium`, and `advanced` for difficulty levels.

vhoof: I also think the user's password_confirmation field doesn't exist anymore. Probably don't have to set created_at/updated_at either, though we may want to set confirmed_at.

PR #201, opened 2019-03-26
search button and error message added on diffrent branch with changes…

jsmith2890:
… requested made because of adding eslint from staging

### Checklist For Submitter

* [ x ] I have documented all new functionality
* [ x ] I have screenshotted new UI changes and attached them to this PR

### DIFF
--- a/frontend/src/components/MainSearch/MainSearch.js
+++ b/frontend/src/components/MainSearch/MainSearch.js
@@ -6,6 +6,7 @@ class MainSearch extends Component {
     super(props);
     this.state = {
       searchTerm: '',
+      error: '',
     };
 
     this.handleChange = this.handleChange.bind(this);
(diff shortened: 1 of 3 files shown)


mglavey:
Hi Jessica,

Looks good, per our irl conversation. Curious as to @vhoof's thoughts re: error handling.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -16,24 +17,52 @@ class MainSearch extends Component {
     this.setState({ searchTerm: event.target.value });
   }
 
+  handleValidation() {
+    let error = '';
+    let isValid = true;
+    const { searchTerm } = this.state;
+    if (!searchTerm) {
```
vhoof: I'd probably condense these two lines into `if (!this.state.searchTerm) {`, since we're not using `searchTerm` anywhere else.

H12:
Mmm yeah good catch @vhoof.

@jsmith2890, can you try deleting your `package-lock.json` file and re-running `npm install`?

frontend/src/components/MainSearch/MainSearch.js
H12: The linter is currently set up to require destructuring state and props.

I'm not super tied to this particular rule, but I do think it aids consistency & readability across the codebase. It also helps with refactoring, since if we start using `searchTerm` elsewhere, or turn `MainSearch` into a stateless component at some point, we only have to deal with one occurrence of `this.state`.

vhoof: Ahh, that's fair. I don't really see any reason to change the rule, it just wasn't one I was familiar with.

PR #207, opened 2019-03-29
Audit CSS

H12:
### Ticket Number:
182

This ticket is responsible for improving our css across the board, but this PR only adds basic linting and fixes the low-hanging fruit.

### Checklist For Submitter

### DIFF
--- a/frontend/.stylelintrc
+++ b/frontend/.stylelintrc
@@ -17,6 +17,6 @@
       "except": ["first-nested"],
       "ignore": ["after-comment"]
     } ],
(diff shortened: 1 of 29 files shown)


H12:
@vhoof I think this is good to go.

This PR only really adds linting rules to enforce syntax stuff. I'll have another PR where I intend to add some more intense rules, like removing `vh` & `vw` from the unit whitelist, then carefully modifying the existing CSS to follow those new rules.

Figure it's probably better to add the CSS linter first, then actually move into the meat-and-potatoes of this css audit in a separate PR.

PR #210, opened 2019-04-04
Feature/8/sign in

KatieLars:
Generating a jwt token to sign in users and added a button to the RLC logo in the login header that links back to homepage.
Please note that at this stage, signing up does not automatically sign in a user.

### Checklist For Submitter

* [ - ] I have documented all new functionality
* [ - ] I have screenshotted new UI changes and attached them to this PR


config/initializers/devise.rb
```diff
@@ -294,6 +294,12 @@
 
   config.jwt do |jwt|
     jwt.secret = ENV['DEVISE_JWT_SECRET_KEY']
+    jwt.dispatch_requests = [
```
vhoof: I think we can remove these, since we figured out the problem was that CORS issue?

frontend/src/components/SignIn/SignIn.js
```diff
@@ -16,41 +22,57 @@ class SignIn extends Component {
...
+  handleChange = (event) => {
+    this.setState({ [event.target.name]: event.target.value });
+  }
+
+  handleSubmit = (event, changeAuthorizedState) => {
+    event.preventDefault();
+    loginUser(this.state)
+      .then((response) => {
+        sessionStorage.jwt = response.headers.authorization;
+        changeAuthorizedState();
```
vhoof: One thing we may want to do above this is also store the user ID, which will be in the response body. That way we can use it to make other requests later, like getting the user's data.

frontend/src/lib/apiClient.js
```diff
@@ -14,6 +14,30 @@ const getRecipes = (search) => {
...
+});
+
+const loginUser = (data) => {
+  const credentials = { email: data.email, password: data.password };
+  return apiClient.post('api/v1/auth/login', {
+    user: credentials,
+  });
+};
+
+const getUser = email => apiClient.get('api/v1/user', {
```
vhoof: This is one of those situations where the user ID is going to be helpful. A GET request is probably going to ignore whatever's in the body. But if you `GET /api/v1/user/1`, that'll get the user info for whoever's primary key is 1. So instead of passing in the email, we may want to pass in the user_id.

In this case, you'll only be able to get that info if you're sending a token authorized to view that user, so you'll want to send along the JWT with this request, as well. You'd put that in the `Authorization` header.

bin/mkdotenv
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+# !/bin/sh
```
vhoof: Looks like those shebang problems are here, too. Maybe you want to rebase or merge in your other branch?

frontend/src/components/SignUp/SignUp.js
```diff
@@ -15,78 +21,95 @@ class SignUp extends Component {
...
+  }
+
+  handleSubmit = (event, changeCreatedState) => {
+    event.preventDefault();
+    createUser(this.state)
+      .then((response) => {
+        changeCreatedState(response.status);
+      })
+      .catch((error) => {
+        if (error.message.includes('422')) {
```
vhoof: Is the API returning a 422 if the email already exists? That's odd, I'd expect a 409. But if this is the case, it's cool, and we can revisit it if I can get the API to behave normally.

frontend/src/components/SignUp/SignUp.js
```diff
@@ -15,78 +21,95 @@ class SignUp extends Component {
...
+          const parentForm = document.getElementsByTagName('form')[0];
+          const errorDiv = document.createElement('p');
+          errorDiv.setAttribute('class', 'errorMessage sign-up');
+          errorDiv.innerHTML = 'A user with that email already exists! Try signing in.';
+          parentForm.insertAdjacentElement('beforeend', errorDiv);
```
vhoof: I think instead of using vanilla JS here, we could just have a div with all of these attributes already on the page, and store the error message in state/render it from there? Or even use a `renderError` function that checks to see if `this.state.errorMessage` is empty, and if not returns a div with the error message. If you think that's doable on your time table pre-demo, that'd be awesome. Otherwise, we can just do this and come back to it.

frontend/src/lib/apiClient.js
```diff
@@ -14,6 +17,32 @@ const getRecipes = (search) => {
...
+const loginUser = (data) => {
+  const credentials = { email: data.email, password: data.password };
+  return apiClient.post('api/v1/auth/login', {
+    user: credentials,
+  });
+};
+
+const getUser = userId => apiClient.get(`api/v1/user/${userId}`, {
+  headers: {
+    Authorization: `Bearer ${sessionStorage.jwt}`,
```
vhoof: I think the JWT was already coming back from the API with the `Bearer` string?

frontend/src/components/SignUp/SignUp.js
KatieLars: Yeah, it gives a 422. I don't know if that is because of the work around with the confirmation email, or what. It says it's an unprocessable request

frontend/src/lib/apiClient.js
KatieLars: You're absolutely right. fixed.

frontend/src/components/SignUp/SignUp.js
KatieLars: That's what I wanted to do . . . I'll add it to the list of things I have to continue/refactor the work I did.

PR #221, opened 2019-04-16
Feature/60/admin recipe backend

jsmith2890:
### Checklist For Submitter

* [ x ] I have documented all new functionality
* [ x ] I have screenshotted new UI changes and attached them to this PR

Also added new npm package so you need to run npm install in the frontend folder

Screen Shot 2019-04-15 at 7 31 59 PM


app/controllers/api/v1/recipe_controller.rb
```diff
@@ -10,8 +10,14 @@ def index
       @recipes = Recipe.search_by_keyword(params[:search])
     else
       @recipes = Recipe.all
+    
+      recipes_with_user = @recipes.map do |recipe|
+        recipe_attributes = recipe.attributes
+        recipe_attributes[:user_name] = "#{recipe.user.first_name}  #{recipe.user.last_name}"
```
vhoof: So because of the way ActiveRecord (the ORM that Rails uses to access the database) works, this loop is going to cause the n+1 query problem, where we'll perform 1 query to get a list of records, then an additional query for each recipe to fetch the user data. If there are 2 recipes in the DB, we'll do 3 queries. If there are a thousand recipes, we'll do 1001 queries.

You might check out https://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations for a solution.

The other piece is that we should just return the whole user object, and let the front end figure out what to display. Right now, this makes the backend code tightly coupled to what we want to display on the front end, which can become fragile and hard to maintain over time.

vhoof: We should probably include the user when we search, too, to keep the data format produced by this endpoint the same.

config/routes.rb
```diff
@@ -15,7 +15,8 @@
              
   namespace :api do
     namespace :v1 do
-      resources :recipe
+      resources :recipe, :user
```
vhoof: What's the `:user` here for? I think this might interfere with signup by overwriting the user creation route.

frontend/src/lib/apiClient.js
```diff
@@ -14,6 +14,9 @@ const getRecipes = (search) => {
     .then(results => results.data);
 };
 
+const getRecipesAdmin = () => apiClient.get('api/v1/recipe')
+  .then(results => results.data);
```
vhoof: I don't think you need a separate function. `getRecipes` should do what you want, just don't pass in anything.

vhoof:
Also probably want to merge/rebase in changes from staging.

jsmith2890:
Alright I made the changes you requested

app/controllers/api/v1/recipe_controller.rb
```diff
@@ -9,9 +9,8 @@ def index
     if params[:search]
       @recipes = Recipe.search_by_keyword(params[:search])
     else
-      @recipes = Recipe.all
+      @recipes = Recipe.includes(:user).all
+      render json: @recipes, :include => [:user]
     end
-    render json: @recipes
```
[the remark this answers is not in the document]
jsmith2890: fixed the render but the :include is needed without it doesnt append the user to the recipe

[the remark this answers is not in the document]
vhoof: Makes sense!

PR #223, opened 2019-04-19
Feature/182/fix css

H12:
This removes some iffy CSS units from the whitelists, and makes corresponding changes to replace those units with supported ones. For areas that used vh and vw (which are responsive based on the viewport size), those areas will no longer be responsive.

I chose to assume a target page width of 1280px when converting the styles over, so the page should look the same as it used to at that width, but when the screen is smaller, some elements will be larger, the flipside being true when the screen is larger than 1280px.

I also added a nodejs & npm version spec to the package.json, which should help with package-lock.json thrash.

### Checklist For Submitter


H12:
@vhoof Looks like we're green! Wanna take a glance at build-setup.sh and make sure it looks alright? NGL, I'm not sure I fully understand the copypasta I threw in there.

Also, I'm gonna pause on merging this until after the demo to limit the risk of running into css regressions.

vhoof:
I think it looks good! If we want, we could also have npm explicitly upgraded to a specific version, but as long as 5.6.0 does what we need (in CircleCI, at least), I think this is totally fine.

PR #225, opened 2019-04-25
Add search bar to header

rmarro:
Added search bar to the header. Only renders in the header on certain pages (not on main page and not on signup/login page)

### Checklist For Submitter

* [X] I have documented all new functionality
* [X] I have screenshotted new UI changes and attached them to this PR

No search bar:
image

Search bar:
image

### DIFF
--- a/frontend/src/components/Header/Header.js
+++ b/frontend/src/components/Header/Header.js
@@ -4,7 +4,7 @@ import rlcLogo from '../../assets/RLC_logo.png';
 import SearchBar from '../SearchBar/SearchBar';
 import './Header.css';
 
-const Header = ({ history }) => (
+const Header = ({ history, showSearchBar }) => (
   <nav className="navbar navbar-white bg-white fixed-top">
     <a className="navbar-brand" href="/">
       <img
@@ -14,7 +14,7 @@ const Header = ({ history }) => (
       />
     </a>
     <div>
-      <SearchBar />
+      {showSearchBar ? <SearchBar /> : null}
     </div>
     <div className="d-flex justify-content-lg-end">
       <button className="btn navbar-btn btn-lg browseButton" href="/" type="button">
--- a/frontend/src/components/Header/Header.js
+++ b/frontend/src/components/Header/Header.js
@@ -1,12 +1,12 @@
 import React from 'react';
 import { withRouter } from 'react-router-dom';
+import PropTypes from 'prop-types';
 import rlcLogo from '../../assets/RLC_logo.png';
 import SearchBar from '../SearchBar/SearchBar';
 import './Header.css';
 
 const Header = ({ history, showSearchBar }) => (
   <nav className="navbar navbar-white bg-white fixed-top">
-
     <a className="navbar-brand" href="/">
       <img
         src={rlcLogo}
@@ -32,5 +32,13 @@ const Header = ({ history, showSearchBar }) => (
     </div>
   </nav>
 );
+
+Header.propTypes = {
+  showSearchBar: PropTypes.bool
+};
+
+Header.defaultProps = {
+  showSearchBar: false
+};
 
 export default withRouter(Header);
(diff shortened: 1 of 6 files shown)


frontend/src/components/SearchBar/SearchBar.css
```diff
@@ -0,0 +1,63 @@
+.findText {
+  color: white;
+  font-size: 2vw;
```
H12: Would it be possible to do this CSS without using `vw` and `vh`? I'm going to be officially blacklisting these units soon via the linter, and avoiding them now will save some back-tracking.

frontend/src/components/SearchBar/SearchBar.css
```diff
@@ -0,0 +1,63 @@
+.findText {
```
H12: NP (nitpick): Would it be possible to use the `kebab-case` standard here?

frontend/src/components/SearchBar/SearchBar.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  handleChange(event) {
+    this.setState({ searchTerm: event.target.value });
+  }
+
+  handleValidation() {
+    let error = '';
+    let isValid = true;
+    const { searchTerm } = this.state;
```
H12: NP: I like to have `const`s precede `let`s to preserve alphabetical order. I also like to have newlines both before and after `if` statements.

frontend/src/components/Header/Header.js
```diff
@@ -1,18 +1,21 @@
 import React from 'react';
 import { withRouter } from 'react-router-dom';
 import rlcLogo from '../../assets/RLC_logo.png';
+import SearchBar from '../SearchBar/SearchBar';
 import './Header.css';
 
-const Header = ({ history }) => (
+const Header = ({ history, showSearchBar }) => (
   <nav className="navbar navbar-white bg-white fixed-top">
+
```
H12: NP: This newline is unnecessary.

frontend/src/containers/Homepage/Homepage.js
```diff
@@ -29,7 +29,7 @@ class Homepage extends Component {
       <div className="homepage container-fluid ">
         <div className="row">
           <div className="header">
-            <Header />
+            <Header showSearchBar={false} />
```
H12: This _might_ be unnecessary, but now that `Header` as two props, I think it could benefit from using [prop-types](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes).

That way, we could specify that `showSearchBar` be a boolean, and then give it a default value of `false` so we don't need to specify `showSearchBar={ false }` below.

frontend/src/components/SearchBar/SearchBar.css
rmarro: a lot of this I copy/pasted from mainSearch, and just realized I didn't need this so it's gone!

frontend/src/containers/Homepage/Homepage.js
rmarro: Currently we only want it to be false on one page, and we want it to render on all the other ones. So if we set the default to be false, would we then have to specify `showSearchBar={ true }` on all the other pages?

H12: Yes and no... While we do need to explicitly set `showSearchBar` to `true`, `<Header showSearchBar />` is just shorthand for `<Header showSearchBar={ true } />`. So we wouldn't need to change any of the existing code.

rmarro: Ah ok this makes sense thanks! I'll go ahead and add this.

frontend/src/components/Header/Header.js
```diff
@@ -30,4 +33,12 @@ const Header = ({ history }) => (
   </nav>
 );
 
+Header.propTypes = {
+  showSearchBar: PropTypes.bool,
```
H12: NP: Let's also add `history` as a prop-type. No need to set a default, though.

PR #227, opened 2019-04-26
Feature/208

fgonza23:
the Donate button in the Header now redirects to the donation link

### Checklist For Submitter


Gemfile.lock
```diff
@@ -1,77 +1,80 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    actioncable (5.2.2)
```
H12: Since the change in `Header` doesn't add any Gems or npm packages, I don't think we should be updating the Gemfile.lock and package-lock.json unnecessarily.

Let's amend the two commits here to remove these changes, and while we're at it (since they both have the same commit message), squash them into one commit.

PR #228, opened 2019-04-26
Feature/203/ Fix the sizing on the footer, also combined the footer lines

fgonza23:
The footer was larger than it needed to be. I reduced the text by one level according to the design doc. I've also combined the last two lines so it reads "Designed by Fluid Designs. App icons by icons8."

### Checklist For Submitter
Screen Shot 2019-04-25 at 11 53 55 PM


vhoof:
It looks like maybe you have bundler 2.0 installed locally, so when you `bundle install` outside the docker container, it's overwriting Gemfile.lock with a non-backwards-compatible format. I think if you can revert/reset the changes to Gemfile.lock and package-lock.json, this would be good to go.

For the future, you may want to:
```
gem uninstall bundler -v 2.0.1
gem install bundler -v 1.17.3
```

That should prevent this from recurring.

vhoof:
@fgonza23 Going to need to fix the merge conflicts, but otherwise did you want to merge this in time for demo?

PR #232, opened 2019-05-02
Feature/191/ user tab of admin dashboard is connected to database

rmarro:
### Checklist For Submitter


app/controllers/api/v1/user_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Api::V1::UserController < ApplicationController
+
+	def index
+		@users = User.includes(:recipes).all
+		render json: @users, :include => [:recipes]
+	end
+end
```
H12: Looks like this file is using tabs instead of spaces. Let's stick to two spaces as a tab.

Let's also ensure every file ends w/ a newline.

frontend/package-lock.json
```diff
@@ -2983,11 +2983,13 @@
             },
             "balanced-match": {
               "version": "1.0.0",
-              "bundled": true
```
H12: Unless we're changing `frontend/package.json`, let's ignore the automatic changes to `frontend/package-lock.json`. It can be risky to update your dependencies with every commit.

rmarro:
image

app/controllers/api/v1/user_controller.rb
fgonza23: Looks great.
I would agree with Henry that two spaces is conventionally used in Ruby and would keep things consistent.

app/controllers/api/v1/user_controller.rb
```diff
@@ -0,0 +1,7 @@
...
```
vhoof: Can we fix the tabbing in this file to 2 spaces? That's the ruby standard.

frontend/src/containers/AdminPanel/AdminPanel.js
```diff
@@ -27,6 +28,21 @@ class AdminPanel extends Component {
...
     });
+
+    getUsers().then((data) => {
+      this.setState({
+        users: this.createFullName(data),
+      });
+    });
+  }
+
+  createFullName = (data) => {
```
vhoof: Nice!

rmarro:
@vhoof Thanks! Strange, I thought the package-lock.json issue we had solved while doing the rebase together. I just pulled down most recent version of staging, and then in my feature branch did `git checkout staging -- frontend/package-lock.json` and it's not saying any changes have been made that I can commit... I must be doing something wrong?

app/controllers/api/v1/user_controller.rb
rmarro: I just fixed this and pushed changes but it's still showing up as tabs...

rmarro: ok I think it's fixed now!!

config/routes.rb
```diff
@@ -16,6 +16,7 @@
   namespace :api do
     namespace :v1 do
       resources :recipe
+      resources :user
```
H12: NitPick: I think this newline is unnecessary.

config/routes.rb
```diff
@@ -16,6 +16,7 @@
...
```
H12: I think technically, to adhere to rails conventions, we should pluralize this and use `users`. But I think this means we'll also need to pluralize `user_controller.rb` and `API::V1::UsersController`.

The reasoning is that it's kinda weird to hit the `api/v1/user` endpoint and get back an array of users.

config/routes.rb
H12: NitPick: I think this newline is unnecessary.

rmarro: deleted

config/routes.rb
rmarro: pluralized to users throughout

PR #281, opened 2019-05-30
Feature/169

Datguyovrder:
The main task was to implement a chevron-down button right below the search image (magnifying glass). When clicked it would disappear and a panel would appear with the text in the picture. At the bottom right of the panel there is the word "Collapse" that closes the panel and the chevron button reappears.

Screen Shot 2019-05-29 at 8 31 03 PM

Screen Shot 2019-05-29 at 8 32 02 PM

### Checklist For Submitter


Datguyovrder:
need to ask Henry how to fix esliint issue. Docker-compose up works but it wreaks havoc on my sublime and circleci tests

frontend/package.json
```diff
@@ -4,19 +4,21 @@
...
     "react": "^16.8.4",
+    "react-bootstrap": "^1.0.0-beta.8",
     "react-bootstrap-table-next": "^3.0.0",
     "react-bootstrap-table2-editor": "^1.2.3",
     "react-bootstrap-table2-paginator": "^2.0.3",
     "react-dom": "^16.8.4",
     "react-fontawesome": "~1.5.0",
     "react-router-dom": "~4.3.1",
-    "react-scripts": "~2.1.8",
+    "react-scripts": "^3.0.1",
```
vhoof: Is there any particular reason we needed to upgrade this?

frontend/package.json
```diff
@@ -4,19 +4,21 @@
   "private": true,
   "dependencies": {
     "axios": "^0.18.0",
+    "bootstrap": "^4.3.1",
     "prop-types": "^15.7.2",
     "react": "^16.8.4",
+    "react-bootstrap": "^1.0.0-beta.8",
```
vhoof: We actually have reactstrap installed further down in the dependencies, which also provides both bootstrap 4 and a react interface. We probably shouldn't use both of them.

frontend/src/components/MainSearch/MainSearch.css
```diff
@@ -80,3 +80,43 @@
...
+  background-color: grey;
+  height: 30px;
+  width: 50px;
+}
+
+.chevron {
+  margin-top: -12px
+}
+
+.content{
```
vhoof: CSS is shared across all the components, so we generally want to use component-specific, descriptive names as opposed to super-generic ones like `content` and `button-box` and `close`.

frontend/src/components/MainSearch/MainSearch.css
```diff
@@ -80,3 +80,43 @@
   font-weight: 500;
   margin: 0px;
 }
+
+.absoluteContainer {
+  /*display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  height: 100px;*/
```
vhoof: If we're not actually using this CSS, we probably don't need it. Is the div even necessary?

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -49,18 +55,43 @@ class MainSearch extends Component {
...
+                          className="chevron"
+                        />
+                      </div>
+                    ) : null}
+                  </div>
+                  {/* ternary operator shortcut (if/else) */}
+                  {this.state.open ? (
+                    <div className="content">
+                      <p> You can search recipe names or ingredients by keyword. Enter your keywords and click the search button or press enter to search. If you use multiple keywords, the search will return only results containing all of your keywords.</p>
+                      <div onClick={(event) => this.togglePanel(event)}>
```
vhoof: This could just be `onClick={this.togglePanel}`.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -49,18 +55,43 @@ class MainSearch extends Component {
...
+                        <img
+                          src="https://img.icons8.com/ios/26/000000/chevron-down.png"
+                          alt="chevron button"
+                          className="chevron"
+                        />
+                      </div>
+                    ) : null}
+                  </div>
+                  {/* ternary operator shortcut (if/else) */}
+                  {this.state.open ? (
```
vhoof: I imagine this is causing a linter error. You'll want to do something like `const { open } = this.state` at the top of the function, and use `open` instead of accessing state directly here.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -49,18 +55,43 @@ class MainSearch extends Component {
...
+                    <button type="submit" className="search-button">
+                      <img
+                        src="https://img.icons8.com/ios/30/000000/search.png"
+                        alt="search"
+                      />
+                    </button>
+                  </div>
+                  <div className="buttonContainer">
+                    {!this.state.open ? (
+                      <div onClick={(event) => this.togglePanel(event)} className="button-box">
```
vhoof: This could just be `onClick={this.togglePanel}`.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -38,6 +40,10 @@ class MainSearch extends Component {
     }
   }
 
+  togglePanel(event) {
```
vhoof: If we're not going to use the event for anything, we don't really need it in the arguments.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -49,18 +55,43 @@ class MainSearch extends Component {
...
+                  {/* ternary operator shortcut (if/else) */}
+                  {this.state.open ? (
+                    <div className="content">
+                      <p> You can search recipe names or ingredients by keyword. Enter your keywords and click the search button or press enter to search. If you use multiple keywords, the search will return only results containing all of your keywords.</p>
+                      <div onClick={(event) => this.togglePanel(event)}>
+                        <p className="close">Collapse</p>
+                      </div>
+                    </div>
+
+                  ) : null}
```
vhoof: How about instead of having two separate ternaries, you just combine them?

Even better, it might be worth pulling this out into a separate function, like `renderExplainerPanel` instead of having a giant ternary in the middle of your render block. Then you can just use a real if statement.

package.json
```diff
@@ -10,5 +10,14 @@
...
     "start": "cd frontend && npm start"
+  },
+  "dependencies": {
+    "install": "^0.12.2",
+    "prop-types": "^15.7.2",
+    "react-accessible-accordion": "^3.0.0",
+    "react-accordeon": "^1.1.1",
+    "react-collapse": "^4.0.3",
+    "react-collapsible": "^2.6.0",
+    "eslint": "^5.16.0"
```
vhoof: We don't want to install dependencies at this level, so probably want to back these changes out. Everything should go in the frontend/package.json file, and these aren't getting used at all.

frontend/package.json
Datguyovrder: should I delete the that specific line from this package.json file?

vhoof: Yeah, I'd remove that one and the bootstrap line further up, since that should be provided by reactstrap.

vhoof: Actually, bootstrap may still be necessary, I'm just lying.

vhoof: OK, I changed my mind again after doing some research. lol We shouldn't need to specifically add bootstrap here.

frontend/src/components/MainSearch/MainSearch.js
Datguyovrder: So if I change togglepanel to:

togglePanel() {
     const { open } = this.state;
}

I can use:

{this.open ? (div) : null}

for the ternary statements?

Datguyovrder:
I made the changes and finally the circleci tests pass

frontend/src/components/MainSearch/MainSearch.js
vhoof: I think this change is worth making. Right now, it's pretty much:

```
if x
  do some stuff

if !x
  do some other stuff
```

Which should probably be more like:

```
if x
  do some stuff
else
  do some other stuff
```

Same thing with these ternaries: `x ? do some stuff : do some other stuff`. But either way, it's going to be more readable breaking it out into a separate function. 15-line ternaries are definitely not ideal.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -38,6 +40,56 @@ class MainSearch extends Component {
...
+  }
+
+  // renderExplainerPanel() {
+  //   const { open } = this.state;
+  //   if (!open) {
+  //     return <renderChevronButton />;
+  //   }
+
+  //   return <renderSearchInstructions />;
+  // }
```
vhoof: I think if you take the `open` handling out of the individual methods, this approach should be fine. But these functions aren't components, so you'll need to `return renderChevronButton();` instead. I'd also flip these so you're checking `open` instead of `!open`.

vhoof: You could even replace this function with a ternary on line 120, like `{ open ? renderSearchInstructions() : renderChevronButton() }`. Since you've reduced it to one line, a ternary makes sense.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -49,18 +101,40 @@ class MainSearch extends Component {
...
+                        className="chevron-image"
+                      />
+                    </div>
+                  </div> */}
+                  { /* <div className="searchInstrucitons">
+                      <p> You can search recipe names or ingredients by keyword. Enter your keywords and click the search button or press enter to search. If you use multiple keywords, the search will return only results containing all of your keywords.</p>
+                      <div onClick={this.togglePanel}>
+                        <p className="close">Collapse</p>
+                      </div>
+                    </div> */}
```
vhoof: Just make sure you delete this commented stuff when you're done.

frontend/src/components/MainSearch/MainSearch.css
```diff
@@ -80,3 +80,33 @@
...
+  background-color: grey;
+  height: 30px;
+  width: 50px;
+}
+
+.chevron-image {
+  margin-top: -12px
+}
+
+.searchInstrucitons{
```
vhoof: *Instructions

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -38,6 +40,56 @@ class MainSearch extends Component {
...
+
+    return null;
+  }
+
+  renderSearchInstructions() {
+    const { open } = this.state;
+
+    if (!open) {
+      return (
+        <div className="searchInstrucitons">
```
vhoof: *Instructions

PR #285, opened 2019-06-06
Feature/234/redirect

andreaMT15:
### Ticket Number:
234

Search bar component must redirect to home page and populate correct search results. 
Screen Shot 2019-06-06 at 7 01 42 PM

### Checklist For Submitter

* [x ] I have documented all new functionality
* [x ] I have screenshotted new UI changes and attached them to this PR


frontend/src/containers/Homepage/Homepage.js
```diff
@@ -9,10 +9,12 @@ import './Homepage.css';
...
-    this.state = {
-      searchTerm: null,
-    };
-
+    const { location } = this.props;
+    if (location.state) {
+      this.state = { searchTerm: location.state.searchTerm };
+    } else {
+      this.state = { searchTerm: null };
+    }
```
SamDesai333: Everything looks good to me!

frontend/src/components/SearchBar/SearchBar.js
```diff
@@ -14,14 +14,24 @@ class SearchBar extends Component {
     this.handleSubmit = this.handleSubmit.bind(this);
   }
 
-  handleChange(event) {
-    this.setState({ searchTerm: event.target.value });
+  setRedirect = () => {
+    const { history } = this.props;
+    const { searchTerm } = this.state;
+    const queryString = `/?q=${searchTerm}`;
+    history.push(queryString, { searchTerm });
```
vhoof: Instead of passing this with location state, it might be preferable to just change the URL and pull the search term directly from the `this.props.location.search` on the Homepage (instead of `this.props.location.state`). That means that if a user types in the a URL with a search term, it'll work the way we expect.

PR #291, opened 2019-06-11
Added CSS for <a> to limit linking to the logo

nickharriscodes:
Limiting the logo link on the header of the login page so it doesn't take up the entire header. 

### Checklist For Submitter

* [-] I have documented all new functionality
* [-] I have screenshotted new UI changes and attached them to this PR


frontend/src/components/LogInPageHeader/LogInPageHeader.css
```diff
@@ -1,4 +1,13 @@
 .logoCenter {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+  width: 108%;
+  padding-bottom: 012px;
+  padding-top: 012px;
+}
+
+a{
```
vhoof: This is clever, but I think there's a cleaner solution. The problem here is ultimately that the logo is displayed as a block, so the whole block is part of the anchor.

But if we just remove `display: block` and the margins, and instead apply the Bootstrap class `text-center` to the enclosing div, that should limit the size of the `a` element and still leave the logo centered.

The other thing is that CSS is shared by all the components on the site, so by styling `a` here, it will change the styling for every anchor tag on the site. Probably something we want to avoid. :)

nickharriscodes: Thanks for clarifying. I figured the code within LoginPageHeader.css would only affect the styling in the login page header - I am learning SO much about React!! The latest commit should correct everything.

PR #295, opened 2019-06-18
Feature/251/admin edit user

rmarro:
Created the edit user modal on the admin page, which saves changes to user information to the database

### Checklist For Submitter


frontend/src/components/AdminControls/Users.js
```diff
@@ -1,133 +1,172 @@
...
-        textAlign: 'center',
-        borderTop: 'none',
-        borderBottom: 'none',
-      },
-      formatter: (cell) => {
-        let dateObj = cell;
-        if (typeof cell !== 'object') {
-          dateObj = new Date(cell);
-        }
+class Users extends Component {
```
vhoof: I mentioned this in person, but curious if we can manage this without changing this to a class-based component. I don't think it's a huge deal, but ideally we'd be moving in the opposite direction.

I think some of the stuff we're doing with state here either wouldn't need to be state or could be refactored pretty easily with the `useState` hook.

frontend/src/components/AdminControls/Users.js
```diff
@@ -1,133 +1,172 @@
...
+
+  toggleEditModal() {
+    this.setState(prevState => ({
+      editModal: !prevState.editModal,
+    }));
+  }
+
+  render() {
+    const { users } = this.props;
+    const columns = [
```
vhoof: I know you're just moving these things around for the most part, but I wonder if we could break the column definitions into a different file or something in this same directory, and then pull them in here?

rmarro: on hold based on our conversation since there are functions in the column definitions

rmarro:
image

frontend/src/components/AdminControls/Users.js
```diff
@@ -1,10 +1,18 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useState } from 'react';
 import BootstrapTable from 'react-bootstrap-table-next';
 import paginationFactory from 'react-bootstrap-table2-paginator';
 import { Type } from 'react-bootstrap-table2-editor';
 import './AdminTables.css';
+import AdminEditModal from '../AdminEditModal/AdminEditModal';
+
+const Users = ({ users, refreshUsers }) => {
+  const [editModal, setEditModal] = useState(false);
```
H12: 😎

frontend/src/components/AdminEditModal/AdminEditModal.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  render() {
+    const {
+      editModal, toggleEditModal, selectedUser,
+    } = this.props;
+    const { alertVisible } = this.state;
+
+    const closeBtn = <button type="button" className="edit-user-close-button" onClick={() => { toggleEditModal(); this.onDismiss(); }}><img src="https://img.icons8.com/windows/32/9b9b9b/cancel.png" alt="close" /></button>;
```
H12: Rather than one-lining that whole onClick, let's abstract the onClick out to an independent function and put the `img` tag on its own line between the button tags.

H12:
Just one small nitpick, but other than that I think it's g2g!

PR #298, opened 2019-06-24
Feature/235/edit search

### LINKED ISSUE #235: As a USER or a VISITOR, I want to be able to see and edit my search so that I can find the recipes I'm looking for
The design doc includes a feature that allows users to see how many search results their search returned and the keywords in their search. They can then "x out" keywords in order to make their search more expansive; removing a keyword should re-execute the search with the remaining search terms.

In the example shown, this should be the result if a user searches for "macaroni cheese truffle". If they remove truffle, the search would re-execute on "macaroni cheese". Note that this means removing a search term should result in the same number of results or more search results.

Acceptance criteria
[ ] When a search is executed, the total number of results are displayed
[ ] When a search is executed, each search term appears in a "lozenge" under the search results count
[ ] If the user clicks x in a lozenge to remove the term, the search is executed with the remaining search terms
[ ] If the user removes a search term, the count of results should be updated to display the count of results in the new query

SamDesai333:
### Ticket Number:
Closes #235

Adds lozenge bubbles to the homepage with the search terms split up.

Screen Shot 2019-06-24 at 6 34 58 PM

### Checklist For Submitter

* [ x] I have documented all new functionality
* [x ] I have screenshotted new UI changes and attached them to this PR


frontend/src/components/SearchLozenge/SearchLozenge.js
```diff
@@ -0,0 +1,28 @@
...
+  }
+  const lozenges = searchTerm.split(' ');
+  return (
+    <div className="searchLozenge container">
+      <div className="row">
+        <div className="lozengeBubbles">
+          {lozenges.map(lozenge => (
+            <div className="lozenges">
+              {lozenge.charAt(0).toUpperCase() }
+              {lozenge.substr(1)}
```
vhoof: I think maybe creating a capitalize function and calling it here would be clearer, like
```
const capitalize = (s) => {
  return s.charAt(0).toUpperCase() + s.slice(1).toLowerCase();
}
```

frontend/src/components/SearchLozenge/SearchLozenge.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import './SearchLozenge.css';
+
+const SearchLozenge = (props) => {
+  let { searchTerm } = props;
+  if (!searchTerm) {
+    searchTerm = '';
+  }
```
vhoof: Maybe `searchTerm = searchTerm || '';` in place of this if condition?

frontend/src/components/SearchLozenge/SearchLozenge.js
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <div className="searchLozenge container">
+      <div className="row">
+        <div className="lozengeBubbles">
+          {lozenges.map(lozenge => (
+            <div className="lozenges">
+              {lozenge.charAt(0).toUpperCase() }
+              {lozenge.substr(1)}
+              {' '}
+              {' x'}
```
vhoof: I assume you're working on it, but definitely using a link or button or something here is gonna be worthwhile.

frontend/src/components/SearchLozenge/SearchLozenge.js
```diff
@@ -0,0 +1,28 @@
...
+  if (!searchTerm) {
+    searchTerm = '';
+  }
+  const lozenges = searchTerm.split(' ');
+  return (
+    <div className="searchLozenge container">
+      <div className="row">
+        <div className="lozengeBubbles">
+          {lozenges.map(lozenge => (
+            <div className="lozenges">
```
vhoof: Maybe just singular "lozenge" as the class name, since there's only one per div?

frontend/src/components/SearchLozenge/SearchLozenge.js
H12: Could also use [`PropTypes`](https://reactjs.org/docs/typechecking-with-proptypes.html) to specify the default. (That would be my preference).

frontend/src/containers/Homepage/Homepage.js
```diff
@@ -31,6 +33,12 @@ const Homepage = (props) => {
               <h5>Based on this week&apos;s popular ingredient donations</h5>
             </div>
           </div>
+          <div className="row">
+            <SearchResults searchTerm={searchTerm} />
+          </div>
+          <div className="row">
+            <SearchLozenge searchTerm={searchTerm} />
+          </div>
```
vhoof: Maybe we should consider embedding `SearchResults` and `SearchLozenge` into `RecipeSearchList`. That way we only have to fetch the search results once.

frontend/src/containers/Homepage/Homepage.js
```diff
@@ -31,6 +33,12 @@ const Homepage = (props) => {
               <h5>Based on this week&apos;s popular ingredient donations</h5>
             </div>
           </div>
+          <div className="row">
+            <SearchResults searchTerm={searchTerm} />
```
vhoof: Maybe we should call this something like `ResultCounter`? `SearchResults` is a little ambiguous.

frontend/src/components/SearchLozenge/SearchLozenge.css
```diff
@@ -0,0 +1,17 @@
+.lozengeBubbles{
+    text-align: center;
+}
+
+
```
vhoof: There's an extra newline here.

frontend/src/containers/Homepage/Homepage.js
```diff
@@ -5,7 +5,9 @@ import Footer from '../../components/Footer/Footer';
 import MainSearch from '../../components/MainSearch/MainSearch';
 import RecipeSearchList from '../../components/RecipeSearchList/RecipeSearchList';
 import BottomSignUp from '../../components/BottomSignUp/BottomSignUp';
+import SearchLozenge from '../../components/SearchLozenge/SearchLozenge';
 import './Homepage.css';
+import SearchResults from '../../components/SearchResults/SearchResults';
```
vhoof: I would put the CSS import at the bottom so the component imports are grouped together.

vhoof:
The only really big thing I might change is better positioning/handling of the `x` so that's an actual delete button. But perhaps it's OK as a first pass?

H12:
@SamDesai333  When you're done implementing @vhoof's feedback, please remove the "Changes Requested" label and `@` me in a comment so I know when it's ready for another look 👍

frontend/src/components/ResultsCounter/ResultsCounter.js
```diff
@@ -0,0 +1,62 @@
...
+        this.setState({
+          recipesFound: data,
+        });
+      });
+    if (!searchTerm) {
+      this.setState({
+        searchTerm: false,
+      });
+    }
+  }
```
vhoof: If we moved this into the `RecipeSearchList` component, we could just pass in the list of recipes (or the length) and remove all of these lifecycle methods/`recipeData`.

frontend/src/components/ResultsCounter/ResultsCounter.js
```diff
@@ -0,0 +1,62 @@
...
+
+  render() {
+    const { searchTerm } = this.state;
+    const { recipesFound } = this.state;
```
vhoof: You can pull multiple things from state at once, like `const { searchTerm, recipesFound } = this.state;`.

frontend/src/components/SearchLozenge/SearchLozenge.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import './SearchLozenge.css';
+
+const SearchLozenge = (props) => {
+  let { searchTerm } = props;
+  searchTerm = searchTerm || '';
```
vhoof: I think @H12's suggestion was to use [`PropTypes`/`defaultProps`](https://reactjs.org/docs/typechecking-with-proptypes.html) to make sure this is set to a default value, instead of this `||` thing.

frontend/src/components/SearchLozenge/SearchLozenge.css
```diff
@@ -0,0 +1,23 @@
...
+    color: #757575;
+    font-size: 20px;
+    box-shadow: 1.5px 2px 1.5px 2px #E7E7E7;
+    margin: 10px;
+    padding-left: 20px;
+    padding-right: 20px;
+    border-radius: 50px 50px;
+}
+
+button{
```
vhoof: CSS is shared across all the pages, so if you style `button` here, it will style buttons across the entire site. We probably don't want to do that, so you'll want to use a class to handle this.

I'd also probably try to position this over toward the right edge of the lozenge, using border-radius to make sure it doesn't escape the container.

frontend/src/components/SearchLozenge/SearchLozenge.js
SamDesai333: Everytime I add propTypes to set the default value, it seems to break my code.

frontend/src/containers/Homepage/Homepage.js
SamDesai333: Changes made!

SamDesai333:
> The only really big thing I might change is better positioning/handling of the `x` so that's an actual delete button. But perhaps it's OK as a first pass?

SamDesai333:
> > The only really big thing I might change is better positioning/handling of the `x` so that's an actual delete button. But perhaps it's OK as a first pass?

Suggestion fixed!

frontend/src/components/SearchLozenge/SearchLozenge.css
SamDesai333: Issue fixed!

frontend/src/components/ResultsCounter/ResultsCounter.js
SamDesai333: Issue fixed!

frontend/src/components/ResultsCounter/ResultsCounter.js
```diff
@@ -0,0 +1,62 @@
+import React, { Component } from 'react';
+import { getRecipes } from '../../lib/apiClient';
+import './ResultsCounter.css';
+
+
+class SearchResults extends Component {
```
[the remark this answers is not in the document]
SamDesai333: Fixed!

PR #300, opened 2019-06-25
Feature/246/password reset

andreaMT15:
### Checklist For Submitter


config/environments/production.rb
```diff
@@ -1,6 +1,17 @@
...
 
+  config.action_mailer.smtp_settings = {
+    address: "smtp.gmail.com",
+    port: 587,
+    domain: "example.com",
+    authentication: "plain",
+    enable_starttls_auto: true,
+    user_name: ENV["GMAIL_USERNAME"],
+    password:ENV["GMAIL_PASSWORD"]
+  }
```
vhoof: I think we'll want to use Sendgrid for this kind of stuff, especially in production. In particular, we'll want to use `SENDGRID_USERNAME` and `SENDGRIND_PASSWORD`. We'll also want to `smtp.sendgrid.net` as the address, and something like `rescuingleftovercuisine.org` as the domain.

I think we should do this in staging, too, and we can sort out how to test it with .env.

Also, probably want a space after `password:`.

config/environments/production.rb
```diff
@@ -1,6 +1,17 @@
...
+
+  config.action_mailer.default_url_options = { :host => 'yourhostsite.herokuapp.com' }
```
vhoof: We'll probably want this to be `ENV['CLIENT_URL']`.

config/environments/development.rb
```diff
@@ -4,6 +4,18 @@
...
+  config.action_mailer.smtp_settings = {
+    address: "smtp.gmail.com",
+    port: 587,
+    domain: "example.com",
+    authentication: "plain",
+    enable_starttls_auto: true,
+    user_name: ENV["GMAIL_USERNAME"],
+    password:ENV["GMAIL_PASSWORD"]
+  }
+config.action_mailer.default_url_options = { :host => 'localhost:3000' }
```
vhoof: We'll want to tab this one over once, and we want to use the front-end, so port 5000.

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,14 +27,43 @@ class NewPassword extends Component {
...
+          changeCreatedState(response.status);
+          history.push(queryString);
+        })
+        .catch((error) => {
+          if (error.message.includes('422')) {
+            const parentForm = document.getElementsByTagName('form')[0];
+            const errorDiv = document.createElement('p');
+            errorDiv.setAttribute('class', 'errorMessage reset-password');
+            errorDiv.innerHTML = 'Please enter a valid password.';
+            parentForm.insertAdjacentElement('beforeend', errorDiv);
```
vhoof: If we're going to copy this from that other page, we should probably use the updated version from SignIn to avoid repeated errors.

vhoof: See https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/staging/frontend/src/components/SignIn/SignIn.js#L37-L40 and https://github.com/the-difference-engine/rescue-leftover-cuisine2/blob/staging/frontend/src/components/SignIn/SignIn.js#L97

frontend/src/containers/ResetPassword/ResetPasswordPage.js
```diff
@@ -1,23 +1,30 @@
...
-    </div>
-    <div className="passwordBackgroundImage">
-      <div className="newPasswordCardWrapper">
-        <NewPassword />
-        <div className="loginFooter row">
-          <Footer />
+const ResetPasswordPage = (props) => {
+  const { location } = props;
+  const parsed = querystring.parse(location.search.slice(1));
+  const token = parsed.q;
```
vhoof: I'd probably use `token` as the query param, though we'll need to see what rails wants to do.

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -48,13 +82,12 @@ class NewPassword extends Component {
...
           <div className="passwordConfirm form-group row">
             <input
               type={isPasswordVisible ? 'text' : 'password'}
               id="inputSignInPasswordConfirm"
               className="password-confirm-input fullWidth form-control-lg"
-              name="confirm-password"
+              name="confirmPassword"
               required
               placeholder="Confirm Password"
               onFocus={event => event.target.setAttribute('placeholder', 'confirmPassword')}
```
vhoof: Can we change this placeholder to say 'Confirm Password'?

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,14 +27,43 @@ class NewPassword extends Component {
...
+            errorDiv.innerHTML = 'Please enter a valid password.';
+            parentForm.insertAdjacentElement('beforeend', errorDiv);
+          }
+        });
+    }
+  };
+
   render() {
-    const { changeAuthorizedState } = this.props;
+    const { changeCreatedState } = this.props;
```
vhoof: Doesn't look like this is being passed in, where does this come from/do?

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,14 +27,43 @@ class NewPassword extends Component {
...
+    const { password, confirmPassword } = this.state;
+    const { resetPasswordToken } = this.state;
+    const queryString = `resetpassword/?q=${resetPasswordToken}`;
+    if (password !== confirmPassword) {
+      alert("Passwords don't match"); // eslint-disable-line no-undef
+    } else {
+      resetPassword(this.state)
+        .then((response) => {
+          changeCreatedState(response.status);
+          history.push(queryString);
```
vhoof: I don't think we want to be redirecting from here to the reset password page, as that's where we are. We could just redirect to the homepage?

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,14 +27,43 @@ class NewPassword extends Component {
     this.setState({ [event.target.name]: event.target.value });
   };
 
+  handleSubmit = (event, changeCreatedState) => {
+    event.preventDefault();
+    const { history } = this.props;
+    const { password, confirmPassword } = this.state;
+    const { resetPasswordToken } = this.state;
+    const queryString = `resetpassword/?q=${resetPasswordToken}`;
```
vhoof: I don't think we need this.

frontend/src/containers/ResetPassword/ResetPasswordPage.js
```diff
@@ -1,23 +1,30 @@
...
-    </div>
-    <div className="passwordBackgroundImage">
-      <div className="newPasswordCardWrapper">
-        <NewPassword />
-        <div className="loginFooter row">
-          <Footer />
+const ResetPasswordPage = (props) => {
+  const { location } = props;
+  const parsed = querystring.parse(location.search.slice(1));
+  const token = parsed.t;
```
andreaMT15: @vhoof in your previous feedback you said to use token as the query string. I tried changing line 12 to const token = token and const token = parsed.token. But I got errors with both of those. So I changed it to parsed.t.  Not sure if I understood your feedback on this correctly.

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,14 +27,34 @@ class NewPassword extends Component {
...
+        .then(this.handleRedirect)
+        .catch((error) => {
+          this.setState({ error: error.message });
+        });
+    }
+  };
+
+  handleRedirect = () => {
+    const { history } = this.props;
+    history.push('/login');
```
vhoof: How about we just cut out this function and move the `history.push` up into the `handleSubmit` function. We can pull the `history` prop at the same time as the `token` prop on line 31.

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,14 +27,34 @@ class NewPassword extends Component {
...
+    const { token } = this.props;
+    event.preventDefault();
+    const { password, confirmPassword } = this.state;
+    if (password !== confirmPassword) {
+      this.setState({ error: "Passwords don't match" });
+    } else {
+      resetPassword(token, password)
+        .then(this.handleRedirect)
+        .catch((error) => {
+          this.setState({ error: error.message });
```
vhoof: I think the error is actually going to be on something like `error.response.data.errors[0].message`. `error.message` is probably just going to be "the request failed with XXX status." I think we may want to add some handling here, like if the error is a 400-level error, it indicates a bad token or bad password.

frontend/src/containers/ResetPassword/ResetPasswordPage.js
```diff
@@ -1,23 +1,31 @@
...
-    <div className="passwordBackgroundImage">
-      <div className="newPasswordCardWrapper">
-        <NewPassword />
-        <div className="loginFooter row">
-          <Footer />
+const ResetPasswordPage = (props) => {
+  const { location } = props;
+  const parsed = querystring.parse(location.search.slice(1));
+  const { token } = parsed;
+  console.log(token);
```
vhoof: We probably don't want to log this. :)

frontend/src/containers/ResetRequest/ResetRequest.css
```diff
@@ -0,0 +1,11 @@
+.btn-wrapper {
```
vhoof: Can we make both of class names in here a little more specific to this page?

frontend/src/containers/ResetRequest/ResetRequest.js
```diff
@@ -0,0 +1,70 @@
...
+  handleChange = (event) => {
+    this.setState({ [event.target.name]: event.target.value });
+  };
+
+  handleSubmit = (event) => {
+    event.preventDefault();
+    const { email } = this.state;
+    requestPasswordReset(email)
+      .then(
+        this.setState({
```
vhoof: I think we need an arrow function here, like `() => this.setState`...

frontend/src/lib/apiClient.js
```diff
@@ -9,7 +9,10 @@ const apiClient = axios.create({
   baseURL,
 });
 
-// AUTHENTICATION
```
vhoof: Can we re-sort the functions in here into their categories and replace the category comments? It's hellishly hard to parse when the user and recipe and login functions are all jumbled together.

frontend/src/lib/apiClient.js
```diff
@@ -27,48 +30,70 @@ const loginUser = (data) => {
...
 
-const adminEditUser = (data, userId) => apiClient.patch(`api/v1/users/${userId}`, {
-  first_name: data.firstName.value,
-  last_name: data.lastName.value,
-  email: data.email.value,
-}).then((response) => {
-  console.log(response);
-}).catch((error) => {
-  console.log(error.response.data);
+// Reset Password Function
```
vhoof: I think the function being called `resetPassword` is self-commenting. :)

frontend/src/lib/apiClient.js
```diff
@@ -27,48 +30,70 @@ const loginUser = (data) => {
...
 
-// USERS
-
-const getUsers = () => apiClient.get('api/v1/users')
-  .then(results => results.data);
-
-const getUser = userId => apiClient.get(`api/v1/users/${userId}`, {
-  headers: {
-    Authorization: sessionStorage.jwt,
+// Reset Request
```
vhoof: Same here re: self-commenting.

H12:
@andreaMT15  When you're done implementing @vhoof's feedback, please remove the "Changes Requested" label and `@` me in a comment so I know when it's ready for another look 👍

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,14 +24,36 @@ class NewPassword extends Component {
...
+      resetPassword(token, password)
+        .then((response) => {
+          if (response) {
+            history.push('/login');
+          }
+        })
+        .catch((error) => {
+          if (error) {
+            this.setState({ error: 'Error' });
+            console.log('Error response', error.response.data.errors); // eslint no-console
```
andreaMT15: @vhoof @H12 I felt like leaving it more general to have a console log that gives us the error response was better for multiple types of errors rather than making it specific to the the reset password token error. This way if the error is unrelated to the password reset token, we will still get specific information for that particular error.

andreaMT15:
@H12 this is ready for review again.

frontend/src/lib/apiClient.js
```diff
@@ -9,7 +9,27 @@ const apiClient = axios.create({
   baseURL,
 });
 
-// AUTHENTICATION
```
vhoof: Is there any particular reason the comments were changed and all the functions were moved around? Seems like it'd be easier just to take what's in staging and add the two functions you need.

config/environments/development.rb
```diff
@@ -4,6 +4,18 @@
...
+
+  config.action_mailer.smtp_settings = {
+    address: "smtp.gmail.com",
+    port: 587,
+    domain: "example.com",
+    authentication: "plain",
+    enable_starttls_auto: true,
+    user_name: ENV["GMAIL_USERNAME"],
+    password:ENV["GMAIL_PASSWORD"]
+  }
```
vhoof: Since we're not actually going to have gmail credentials on development machines, maybe we should remove all this configuration?

andreaMT15:
api client file still needs to be fixed to match staging again. Will complete later this evening.

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,17 +24,51 @@ class NewPassword extends Component {
...
+    const { isPasswordVisible, error } = this.state;
 
+    if (error) {
+      return (
+        <div className="errorMessage">
+          <p>
+            {error}
+            {' '}
+            <Link to="/resetrequest" id="requestLink">
+              Click here to request a new password reset token.
```
vhoof: I might say "You can <Link>request a new password reset token</Link>." I think we can rely on most people knowing that you can click on links. :)

frontend/src/components/NewPassword/NewPassword.js
```diff
@@ -22,17 +24,51 @@ class NewPassword extends Component {
...
+      this.setState({ error: "Passwords don't match" });
+    } else {
+      resetPassword(token, password)
+        .then(() => {
+          history.push('/login');
+        })
+        .catch((error) => {
+          if (error.response.data.errors.reset_password_token[0]) {
+            this.setState({
+              error: `Reset password token ${error.response.data.errors.reset_password_token[0]}.`,
```
vhoof: What does this actually end up returning?

app/controllers/api/v1/recipe_controller.rb
```diff
@@ -0,0 +1,16 @@
+class Api::V1::RecipeController < ApplicationController
```
vhoof: Where did this file come from?

frontend/package.json
```diff
@@ -14,7 +14,7 @@
     "react-router-dom": "~4.3.1",
     "react-scripts": "~2.1.8",
     "reactstrap": "^7.1.0",
-    "universal-cookie": "~4.0.2"
+    "universal-cookie": "^4.0.2"
```
vhoof: I think we probably don't want to make changes to package.json/package-lock.json unless we actually changed some packages, so you can reset these changes.

frontend/src/components/NewPassword/NewPassword.js
vhoof: I think we'll also want to handle the case where something goes wrong that isn't token-related. Like if the reset password token thing is there, then yeah, display this error. Else display something like, "Something went wrong with our system. Please try again."

frontend/src/components/NewPassword/NewPassword.js
andreaMT15: @vhoof The reason I worded it that way is because when you look at it on the front end it's not obvious that it's a link. Or at least I didn't think it's obvious since it's the same color text :) I'll change it though. :)

app/controllers/api/v1/recipe_controller.rb
andreaMT15: @vhoof I'm not sure how this made it into our PR. Neither Fransisco or I have made changes to this file. All i can think of is that it got pushed because I merged staging into our branch.

frontend/package.json
andreaMT15: @vhoof we kept getting errors that it couldn't find the universal-cookie package. So when we installed it, is when it made changes to package.json. How do we reset this without getting an error again?

vhoof: Yeah, the package was added in staging. Since it's already in package.json, all you need to do is `docker-compose build` to update your dependencies.

app/controllers/api/v1/recipe_controller.rb
vhoof: We should probably remove it. The actual file is `recipes_controller`, so not sure what this one is. And either way, it shouldn't be in your PR.

frontend/package.json
vhoof: And you'll want to reset these files to their staging state first.

frontend/src/components/NewPassword/NewPassword.js
vhoof: Ahh, I'd probably style it like the other links on the page (that blue or whatever).

andreaMT15:
@vhoof This is finally ready for final review! :) I think we should finally be able to merge this now.

PR #301, opened 2019-06-25
Feature/277

nickharriscodes:
This PR covers the front-end parts of the ticket 277. These features should work visually, but they don't yet make changes to user data on the backend. 

In the admin panel, the icons on the far right will change depending on whether the user is_suspended or not. Styling for this new column may need to be tweaked. When the X-circle button is clicked, a modal pops up prompting the admin-user to suspend the selected user. 

Much of the new code can be found in commit 4089ee4

Screen Shot 2019-06-25 at 2 03 17 PM

### Checklist For Submitter


frontend/src/components/AdminControls/Users.js
```diff
@@ -1,133 +1,216 @@
...
+    this.setState(prevState => ({
+      suspendModal: !prevState.suspendModal,
+    }));
+  }
+
+  render() {
+    const { users } = this.props;
+    const columns = [
+      {
+        dataField: 'inStock', // name?
```
SamDesai333: Everything here seems fine. Maybe take out the comment?

nickharriscodes: Does the comment count as "documentation"?

frontend/src/components/AdminControls/Users.js
```diff
@@ -1,133 +1,216 @@
...
+          textAlign: 'center',
+          borderTop: 'none',
+          borderBottom: 'none',
+        },
+        align: 'left',
+        formatter: (cell) => {
+          if (cell) {
+            return (
+              <div>
+                <button type="button" className="admin-edit-button">
```
rmarro: This one and the one below should probably have a different className (something like admin-suspend-button?) 
Then you could also fix the sizing because I think currently when a user is suspended, the icon shows up as 25x25 and looks a bit small (the others are 32x32)

frontend/src/components/AdminControls/Users.js
```diff
@@ -1,133 +1,216 @@
...
+          </div>
+        ),
       },
+      // Separate column added for suspend-user icons.
+      // Will display one icon or the other based on whether
+      // user is suspended or not.
       {
-        text: 'All',
-        value: users.length,
+        dataField: 'is_suspended',
```
rmarro: I think making this a new column is a creative workaround to the button change challenge. The only issue to resolve is the formatting-- since it's now a new column it's really far from the other button, so may need some changed padding/margin to look like it did before.

frontend/package.json
```diff
@@ -16,7 +16,7 @@
     "reactstrap": "^7.1.0"
   },
   "devDependencies": {
-    "eslint": "5.12.0",
+    "eslint": "^5.12.0",
```
JesseDavid: What's the reason for the update to the `^`? I only ask because I've been bitten in the past by adding the `^` or `~`, and having a package update break my code.

frontend/src/components/AdminControls/Users.js
```diff
@@ -1,133 +1,216 @@
...
+        formatter: (cell) => {
+          let dateObj = cell;
+          if (typeof cell !== 'object') {
+            dateObj = new Date(cell);
+          }
+
+          const day = `0${dateObj.getUTCDate()}`.slice(-2);
+          const month = `0${dateObj.getUTCMonth() + 1}`;
+          const year = dateObj.getUTCFullYear();
+          return `${month}/${day}/${year}`;
```
JesseDavid: 👍

frontend/src/components/AdminEditModal/AdminEditModal.css
```diff
@@ -0,0 +1,45 @@
...
+  margin-top: 125px;
+}
+
+.edit-user-modal .modal-content {
+  background: #f2f2f2;
+  border: none;
+}
+
+.edit-user-modal .modal-header {
+  padding-left: 30%;
```
JesseDavid: Most of the padding I've seen so far is using `px` rather than `%`, is there a specific reason this one works better as a percent? Have you checked at different widths if it looks good?

frontend/src/components/AdminModal/AdminModal.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import {
+  Button, Modal, ModalHeader, ModalBody, ModalFooter, Form, FormGroup, Label, Col, Input,
+} from 'reactstrap';
+// import './AdminModal.css';
```
JesseDavid: I think this comment can be removed.

frontend/package.json
nickharriscodes: I'm not sure... I didn't actually make that change. But that's a good point.

frontend/src/components/AdminEditModal/AdminEditModal.css
nickharriscodes: Good point. Should probably keep it consistent.

PR #302, opened 2019-06-28
Promote staging to master for demo

H12:


H12:
Looks like staging is a-ok

PR #303, opened 2019-07-02
Feature/6

ashzagorski:
Enable ability to navigate to individual recipe page from home page and create recipe show page so individual recipe's information displays on recipe page. 

Screen Shot 2019-07-01 at 8 29 53 PM

Screen Shot 2019-07-01 at 8 30 12 PM
Screen Shot 2019-07-01 at 8 30 23 PM

### Checklist For Submitter

* [ +] I have documented all new functionality
* [+ ] I have screenshotted new UI changes and attached them to this PR


frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -3,26 +3,44 @@ import { withRouter } from 'react-router-dom';
...
+      snippet: '',
+      ingredients: [],
+      difficulty: '',
+      duration: '',
+      servings: '',
+      photo: '',
     };
   }
 
+  componentWillMount() {
```
andreaMT15: According to the React documentation componentWillMount is a legacy method and should be replaced with componentDidMount.

andreaMT15: According the the React documentation componentWillMount is legacy method and considered unsafe. I believe componentDidMount is the life cycle method that should be used instead.

frontend/src/components/ScrollToTop/ScrollToTop.js
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react';
+
+
+class ScrollToTopOnMount extends Component {
```
H12: Does this need to be its own component?

It looks like we could just put the `window.scrollTo` in `Recipes.js`'s `componentDidMount`.

vhoof: This is probably my fault, but :+1: on the change.

frontend/src/containers/RecipePage/Recipe.css
```diff
@@ -32,12 +32,15 @@
   padding: 0px;
 }
 
+.image-wrapper {
```
vhoof: Can we call this something recipe-page-specific?

frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -3,26 +3,44 @@ import { withRouter } from 'react-router-dom';
...
+        title: response.data.title,
+        snippet: response.data.snippet,
+        ingredients: response.data.ingredients,
+        photo: response.data.photos[0],
+        directions: response.data.directions,
+        difficulty: response.data.difficulty,
+        duration: response.data.duration,
+        servings: response.data.servings,
+      });
+    });
```
vhoof: I think we may need some error-handling here, but we could do that on a different ticket.

vhoof:
Somehow the most recent commit has a whoooooole bunch of stuff in the vendor folder that shouldn't be there, and some changes to the mailer/user verification stuff in the backend. My guess is that stuff is coming from a different branch. All of that is going to need to be rolled back.

I really recommend taking a look at what you're actually committing before you commit it.

PR #304, opened 2019-07-11
Feature/277

nickharriscodes:
We have implemented basic suspend-user functionality. On the users screen of the admin panel, the admin can click an "x" next to the users name, triggering a modal with a "Suspend User" button. When the button is clicked, the users data is updated and the icon next to the name changes to indicate the change. When a suspended user attempts to login, they are given a message that says "Your account is suspended. Please contact an administrator for assistance." Suspended users are unable to login. 

Commits made between June 19th - July 9th are the relevant commits to this PR. 

Screen Shot 2019-06-25 at 2 03 17 PM
Screen Shot 2019-07-10 at 8 19 26 PM

### Checklist For Submitter


frontend/src/components/AdminControls/Users.js
```diff
@@ -58,6 +64,7 @@ const Users = ({ users, refreshUsers }) => {
           dateObj = new Date(cell);
         }
 
+
```
SamDesai333: Everything looks good here! Just try and delete some extra spaces if you can!

app/controllers/api/v1/users_controller.rb
```diff
@@ -1,4 +1,4 @@
-class Api::V1::UsersController < ApplicationController
+  class Api::V1::UsersController < ApplicationController
```
vhoof: Think this may have gotten tabbed over accidentally?

app/controllers/api/v1/users_controller.rb
```diff
@@ -9,13 +9,14 @@ def show
     render json: @user
   end
 
+  # fix this to handle updating of the user data
   def update
-    @user = User.find_by!(id: params[:id])
+    @user = User.find_by(id: params[:id])
```
vhoof: I think we want to keep the exclamation mark so that this returns a 404 if the user isn't found.

I also think we may need to do some validation here to make sure `is_suspended` is only updated by admins, but we will have to do that later.

app/controllers/api/v1/users_controller.rb
```diff
@@ -9,13 +9,14 @@ def show
     render json: @user
   end
 
+  # fix this to handle updating of the user data
```
vhoof: Probably want to remove this comma?

app/controllers/api/v1/users_controller.rb
```diff
@@ -9,13 +9,14 @@ def show
...
   end
 
+  # fix this to handle updating of the user data
   def update
-    @user = User.find_by!(id: params[:id])
+    @user = User.find_by(id: params[:id])
     @user.update(user_params)
   end
 
+  # fix this for above (this helps format the data to spit it back into update above)
```
vhoof: This comment, too?

frontend/.eslintrc
```diff
@@ -17,7 +17,8 @@
     "react/prop-types": 0,
     "no-underscore-dangle": 0,
     "import/imports-first": ["error", "absolute-first"],
-    "import/newline-after-import": "error"
+    "import/newline-after-import": "error",
+    "max-len": 0,
```
vhoof: I think we want to continue enforcing line length, so probably want to remove this.

frontend/package.json
```diff
@@ -16,7 +16,7 @@
     "reactstrap": "^7.1.0"
   },
   "devDependencies": {
-    "eslint": "5.12.0",
+    "eslint": "^5.12.0",
```
vhoof: We want eslint to be pinned to this specific version, so we don't want the `^`. Which means we also need to reset the package-lock changes.

frontend/src/components/AdminEditModal/AdminEditModal.js
```diff
@@ -45,16 +45,7 @@ class AdminEditModal extends React.Component {
...
-      toggleEditModal();
-      this.onDismiss();
-    };
-
-    const closeBtn = (
-      <button type="button" className="edit-user-close-button" onClick={handleClose}>
-        <img src="https://img.icons8.com/windows/32/9b9b9b/cancel.png" alt="close" />
-      </button>
-    );
+    const closeBtn = <button type="button" className="edit-user-close-button" onClick={() => { toggleEditModal(); this.onDismiss(); }}><img src="https://img.icons8.com/windows/32/9b9b9b/cancel.png" alt="close" /></button>;
```
vhoof: Multiple lines are more readable.

frontend/src/components/AdminSuspendModal/AdminSuspendModal.js
```diff
@@ -0,0 +1,51 @@
...
+    } = this.props;
+
+    const handleSuspend = () => {
+      suspendUser(selectedUser, selectedUser.id).then(() => {
+        toggleSuspendModal();
+        refreshUsers();
+      });
+    };
+
+    const closeBtn = <button type="button" className="suspend-user-close-button" onClick={toggleSuspendModal}><img src="https://img.icons8.com/windows/32/9b9b9b/cancel.png" alt="close" /></button>;
```
vhoof: I'd also break this into multiple lines, and any text below that causes a linting issue.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -67,11 +67,7 @@ class MainSearch extends Component {
   renderSearchInstructions() {
     return (
       <div className="searchInstructions">
-        <p>
-          You can search recipe names or ingredients by keyword. Enter your keywords and click the
-          search button or press enter to search. If you use multiple keywords, the search will
-          return only results containing all of your keywords.
-        </p>
+        <p> You can search recipe names or ingredients by keyword. Enter your keywords and click the search button or press enter to search. If you use multiple keywords, the search will return only results containing all of your keywords.</p>
```
vhoof: Maybe your editor is changing these automatically? If so, I'd turn off that feature, and keep this as a multi-line paragraph.

frontend/src/components/SignIn/SignIn.js
```diff
@@ -13,9 +13,19 @@ class SignIn extends Component {
       email: '',
       password: '',
       validationErrorText: '',
+      response: [],
+      suspendedUserList: [],
     };
   }
 
+  componentDidMount() {
+    getUsers().then((data) => {
```
vhoof: Once we actually have login working on the site, this page won't have access to the `getUsers` call, as you'll need to be an admin (otherwise we'd basically be publishing every user's email address publicly).

frontend/src/components/SignIn/SignIn.js
```diff
@@ -42,6 +52,27 @@ class SignIn extends Component {
       });
   };
 
+
+  handleSuspendedUser = () => {
```
vhoof: I think we'll actually want to handle this on the backend, so that Devise doesn't allow a user to log in. Maybe we could add an `active_for_authentication?` method to the User model in Rails, which returns false if the user is suspended, like in https://stackoverflow.com/questions/6004216/devise-how-do-i-forbid-certain-users-from-signing-in

frontend/src/components/AdminSuspendModal/AdminSuspendModal.js
```diff
@@ -0,0 +1,51 @@
...
+    };
+  }
+
+  render() {
+    const {
+      toggleSuspendModal, suspendModal, selectedUser, refreshUsers,
+    } = this.props;
+
+    const handleSuspend = () => {
+      suspendUser(selectedUser, selectedUser.id).then(() => {
```
vhoof: Maybe instead of sending the whole user and handling the toggling in the API call, we actually just send the user ID and `true` or `false` to `suspendUser`? That way we can keep the actual toggling logic here where it's being done, and the `suspendUser` can just submit what we tell it to submit.

db/migrate/20190719003405_remove_is_suspended_from_users.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveIsSuspendedFromUsers < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :users, :is_suspended, :boolean
```
vhoof: So since none of these migrations have been merged to the master (or staging) branch, you don't need multiple migrations to change the name of this column. You could just change the first migration, delete the schema.rb file, and start over with a clean database. But why the change? `is_suspended` is probably a better name, as it shows right there in the name that it's a boolean. Also, why remove the default value?

frontend/package-lock.json
```diff
@@ -2768,9 +2768,9 @@
       }
     },
     "callsites": {
```
vhoof: This file probably shouldn't have changes?

frontend/src/lib/apiClient.js
```diff
@@ -46,6 +46,14 @@ const getUser = userId => apiClient.get(`api/v1/users/${userId}`, {
   },
 });
 
+const suspendUser = (data, userId) => apiClient.patch(`api/v1/users/${userId}`, {
+  suspended: !data.suspended,
+});
+
+const getSuspendedUsers = () => apiClient.get('api/v1/users', { is_suspended: true }).then(response => response.data);
```
vhoof: Can probably remove this.

app/models/user.rb
```diff
@@ -16,5 +16,13 @@ def skip_confirmation_method
     self.skip_confirmation!
   end
 
+  def destroy
+    update_attributes(suspended: true) unless suspended
```
vhoof: What're you trying to do here?

config/locales/devise.en.yml
```diff
@@ -8,7 +8,7 @@ en:
       send_paranoid_instructions: "If your email address exists in our database, you will receive an email with instructions for how to confirm your email address in a few minutes."
     failure:
       already_authenticated: "You are already signed in."
-      inactive: "Your account is not activated yet."
+      inactive: "Your account is suspended."
```
vhoof: I think it may be better to handle this in an `inactive_message` method on the user model, as we'll eventually need to support both suspended and not-activated states.

Datguyovrder: This was just us trying to test if devise would show the suspended message when the user is suspended. All we got was the general "check your email or password" 401 message. We will change this back and try to inactive_method in the User model.

app/models/user.rb
Datguyovrder: Instead of the normal destroy method which would delete the user's data, we were tweaking it so that if called upon then suspended would be set to true. We were testing if we could call destroy from the suspend modal button.

frontend/package-lock.json
Datguyovrder: We have no idea what this would be related to?

db/migrate/20190719003405_remove_is_suspended_from_users.rb
Datguyovrder: We tried to follow the guidelines from the file you sent us. We had not idea if the default value for is_suspended would affect the results. Since we rake db:migrate, what would be the best commands to get back to is_suspended?

frontend/src/components/SignIn/SignIn.js
```diff
@@ -34,11 +36,10 @@ class SignIn extends Component {
         changeAuthorizedState(response.data.id);
       })
       .catch((error) => {
-        if (error.message.includes('401')) {
-          this.setState({
-            validationErrorText: 'Please make sure you have the correct email and passwords!',
-          });
-        }
+        console.log(error.response.data.errors[0].message);
```
vhoof: Let's trim this console.log.

frontend/src/components/AdminSuspendModal/AdminSuspendModal.js
vhoof: I'd still love to fix this function call, to something like `suspendUser(selectedUser.id, !selectedUser.is_suspended)`, and change the `suspendUser` function in `apiClient` to just submit the second argument instead of toggling.

frontend/src/lib/apiClient.js
```diff
@@ -46,6 +46,15 @@ const getUser = userId => apiClient.get(`api/v1/users/${userId}`, {
   },
 });
 
+const suspendUser = (data, userId) => apiClient.patch(`api/v1/users/${userId}`, {
+  is_suspended: !data.is_suspended,
+});
+
+// const getSuspendedUsers = () => apiClient.get('api/v1/users', { is_suspended: true }).
+// then(response => response.data);
```
vhoof: Let's remove this. Commented-out code is Bad.

frontend/src/components/SignIn/SignIn.js
```diff
@@ -88,7 +89,7 @@ class SignIn extends Component {
             <a href="##########">I forgot my password</a>
           </div>
           <div className="row">
-            <button className="signInButton signUpButton btn btn-lg btn-block" type="submit" valid>
+            <button className="signInButton signUpButton btn btn-lg btn-block" type="button" onClick={this.handleSubmit} valid>
```
vhoof: Can we make this `valid="true"` so we don't get weird warnings?

frontend/src/lib/apiClient.js
```diff
@@ -46,6 +46,15 @@ const getUser = userId => apiClient.get(`api/v1/users/${userId}`, {
...
+
+// RECIPE
```
vhoof: I think we can also remove this comment, since it's in the wrong place.

frontend/package-lock.json
vhoof: We should probably reset the changes. You can do this with `git fetch origin staging && get checkout origin/staging -- frontend/package-lock.json`.

H12:
@Datguyovrder  When you're done implementing @vhoof's feedback, please remove the "Changes Requested" label and `@` me in a comment so I know when it's ready for another look 👍

frontend/src/components/AdminSuspendModal/AdminSuspendModal.js
```diff
@@ -0,0 +1,68 @@
...
+    const {
+      toggleSuspendModal, suspendModal, selectedUser, refreshUsers,
+    } = this.props;
+
+    // const handleSuspend = () => {
+    //   suspendUser(selectedUser, selectedUser.id).then(() => {
+    //     toggleSuspendModal();
+    //     refreshUsers();
+    //   });
+    // };
```
vhoof: Commented code.

package.json
```diff
@@ -11,5 +11,8 @@
     "lint": "cd frontend && npm run lint",
     "postinstall": "npm run build && npm run deploy",
     "start": "cd frontend && npm start"
+  },
+  "dependencies": {
+    "universal-cookie": "^4.0.2"
```
vhoof: Not sure how this got here, but it shouldn't be. And you'll want to reset changes to the root package-lock.json.

frontend/src/components/AdminSuspendModal/AdminSuspendModal.js
```diff
@@ -0,0 +1,61 @@
...
+} from 'reactstrap';
+import './AdminSuspendModal.css';
+import { suspendUser } from '../../lib/apiClient';
+
+class AdminSuspendModal extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+    };
+  }
```
vhoof: Since we don't have any state, we don't need to override the constructor.

frontend/src/lib/apiClient.js
vhoof: I'd still remove this comment.

db/seeds.rb
```diff
@@ -9,6 +9,8 @@
 
 User.create(email: "bob@domain.com", password: "Bob@RLC123", password_confirmation: "Bob@RLC123", username: "Bob@RLC", first_name: "Bob", last_name: "Domain", profile_photo: "/profilePic.PNG", confirmed_at: Time.now)
 User.create(email: "joe@domain.com", password: "Joe@RLC123", password_confirmation: "Joe@RLC123", username: "Joe@RLC",first_name: "Joe", last_name: "Domain", confirmed_at: Time.now)
+User.create(email: "vespucci@domain.com", password: "Vespucci@RLC123", password_confirmation: "Vespucci@RLC123", username: "Vespucci@RLC",first_name: "Vespucci", last_name: "Domain", confirmed_at: Time.now, is_suspended: true)
+
```
vhoof: There's an extra newline here.

PR #306, opened 2019-07-23
Feature/299

ashzagorski:
### Checklist For Submitter

* [+ ] I have documented all new functionality
* [+ ] I have screenshotted new UI changes and attached them to this PR

Screen Shot 2019-07-22 at 8 34 27 PM
Screen Shot 2019-07-22 at 8 34 42 PM

### DIFF
--- a/frontend/src/components/MainSearch/MainSearch.js
+++ b/frontend/src/components/MainSearch/MainSearch.js
@@ -10,29 +10,26 @@ class MainSearch extends Component {
       searchTerm: '',
       error: '',
       open: true,
-   };
+    };
 
     this.handleChange = this.handleChange.bind(this);
     this.handleSubmit = this.handleSubmit.bind(this);
     this.togglePanel = this.togglePanel.bind(this);
   }
 
-  componentWillUnmount() {
-    const cookies  = new Cookies();
-    cookies.set('visit', 'set', { path: '/' });
-    console.log(cookies.get('visit'));;
+  componentWillMount() {
+    const cookies = new Cookies();
+    if (cookies.get('visit') === 'set') {
+      this.setState({
+        open: false,
+      });
+    }
   }
 
-  componentWillMount() {
-      const cookies = new Cookies();
-      if (cookies.get('visit') === 'set') {
-        console.log('2');
-        this.setState({
-        open: false,
-        });
-      }
-      console.log(this.state.open);
-    }
+  componentWillUnmount() {
+    const cookies = new Cookies();
+    cookies.set('visit', 'set', { path: '/' });
+  }
 
   handleRedirect = () => {
     const { history } = this.props;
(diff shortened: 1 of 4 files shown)


frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -8,14 +9,28 @@ class MainSearch extends Component {
...
+        open: false,
+      });
+    }
+  }
+
+  componentWillUnmount() {
+    const cookies = new Cookies();
+    cookies.set('visit', 'set', { path: '/' });
+  }
+
```
SamDesai333: Everything looks good to me!

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -8,14 +9,28 @@ class MainSearch extends Component {
...
-      open: false,
+      open: true,
     };
 
     this.handleChange = this.handleChange.bind(this);
     this.handleSubmit = this.handleSubmit.bind(this);
     this.togglePanel = this.togglePanel.bind(this);
   }
 
+  componentWillMount() {
```
vhoof: I believe `componentWillMount` is considered unsafe, and we should just use `componentDidMount`. I might consider setting `open` to false by default, and setting it to `true` if the cookie is missing. I might also set the `visit` cookie in `componentDidMount`, as well, after you check it, and then we don't have to worry about `componentWillUnmount`.

H12:
@ashzagorski When you're done implementing @vhoof's feedback, please remove the "Changes Requested" label and `@` me in a comment so I know when it's ready for another look 👍

ashzagorski:
@H12

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -16,6 +17,16 @@ class MainSearch extends Component {
     this.togglePanel = this.togglePanel.bind(this);
   }
 
+  componentDidMount() {
+    const cookies = new Cookies();
+    if (cookies.get('visit') !== 'set') {
+      this.setState({
+        open: true,
+      });
+      cookies.set('visit', 'set', { path: '/' });
```
H12: NitPick: I think we should change the cookie naming a little. Something like `cookies.set('rlc', 'visited', { path: '/' });` would be a little more descriptive.

frontend/src/components/MainSearch/MainSearch.js
```diff
@@ -16,6 +17,16 @@ class MainSearch extends Component {
     this.togglePanel = this.togglePanel.bind(this);
   }
 
+  componentDidMount() {
+    const cookies = new Cookies();
+    if (cookies.get('visit') !== 'set') {
+      this.setState({
+        open: true,
```
H12: NitPick: I think we could be a little more descriptive than just `open`, since that could refer to anything, not just the search instructions. Let's do something like `showSearchInstructions`.

ashzagorski:
@H12 I changed the naming of the cookie, I attempted to change naming of search instruction variable however I received a linting error that the line was over the 100 max, I can only name it something with 4 characters.

PR #309, opened 2019-07-27
Fix toggling of search instructions

vhoof:

I was checking the site to make sure it was ready to demo on Monday, and realized that you could no longer toggle the search instructions. This was due to the state being renamed to `show`, but it wasn't changed in the `togglePanel` method. I fixed that.

I also renamed the state variable to `showInstructions` to be more clear, and renamed the render function to `renderInstructions`, which helped it all fit in one ternary line.


vhoof:
Going to go ahead and merge this so it makes it in before Monday.

PR #313, opened 2019-08-05
Fix auth-related page styling and several other things

vhoof:
Since we first implemented the login page, it's looked a little squished compared to the designs, and at different resolutions it starts to look a little bonkers. Additionally, we keep implementing pages with this format (the same background, a stripped-down header, and 1+ cards with a form). For example, we just implemented the reset password page, and the password reset request page was implemented without being correctly styled.

Additionally, a number of small issues have leaked into various pages, in particular using the wrong colors. I also have come to the conclusion that while the style guide suggests using `Nunito` for the font, this makes the site look subtly bizarre, and it's fairly clear the font in the designs is closer to `Nunito Sans`. I believe this may be one of the errors in the style guide.

In this PR, I switched the font to Nunito Sans. I also tweaked colors on some of the pages that I was already touching to better match the designs.

The main changes on this PR, though, surround the login and password reset pages. These have been tightened up style-wise, and I've built more abstracted components (`AuthPage`, `AuthCard`, and `AuthInput`) to handle the repeatable pieces. I've rewritten the login page, reset password page, password reset request page, sign in, sign up, reset request, and new password components to leverage these building blocks.

Finally, I deleted some duplicate `TagBar` components, and the `RecipeBase` stuff that was leftover from an old version of the site.

Sorry for the size of this PR. The two commits are mostly broken up into 1) the auth pages refactor and 2) the other tweaks, in case it's easier to peruse them separately.


frontend/src/containers/ResetRequestPage/ResetRequestPage.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { AuthPage } from '../../components/AuthItems/AuthItems';
+import ResetRequest from '../../components/ResetRequest/ResetRequest';
+
+const ResetRequestPage = () => (
+  <AuthPage>
+    <ResetRequest />
+  </AuthPage>
+);
+
```
H12: Make it go away

PR #315, opened 2019-08-06
Feature/164 2

Datguyovrder:

After a user logs in, when the user goes to their profile page, they see their name at the top of the page, like in the screenshot below. This works for all current users. 

screen_shot_2019-08-05_at_8 44 28_pm

To make sure it works with other users, another user (Joe Domain) was logged in.

screen_shot_2019-08-05_at_8 51 56_pm

If the user is not logged in then an 401 error message appears on the profile page telling the user to login.

screen_shot_2019-08-05_at_8 44 51_pm

### Checklist For Submitter

### DIFF
--- a/frontend/src/containers/ProfilePage/ProfilePage.js
+++ b/frontend/src/containers/ProfilePage/ProfilePage.js
@@ -89,20 +89,19 @@ class ProfilePage extends Component {
     }
   }
 
-  render() {
+  renderProfile() {
     const { user, setJwt } = this.props;
     const { activeTab, recipes } = this.state;
 
-    return (
-      <div className="profile-page container-fluid">
-        <div className="row">
-          <div className="profile-header">
-            <Header showSearchBar user={user} setJwt={setJwt} />
+    if (user) {
+      return (
+        <div className="profile-page container-fluid">
+          <div className="row">
+            <div className="profile-header">
+              <Header showSearchBar user={user} setJwt={setJwt} />
+            </div>
           </div>
-        </div>
-        {/* Component for top user info and bio section */}
-        { user
-          ? (<div>
+          <div>
             <UserInfo user={user} />
             <div className="row user-panel">
               <div className="col-md-2 user-nav-bar">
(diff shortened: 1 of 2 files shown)


frontend/src/containers/ProfilePage/ProfilePage.js
```diff
@@ -101,7 +101,7 @@ class ProfilePage extends Component {
           </div>
         </div>
         {/* Component for top user info and bio section */}
-        <UserInfo />
+        {user ? <UserInfo user={user} /> : <h1>401 (unauthorized) Login to access this page.</h1>}
```
vhoof: I don't think we want to talk to our users about 401 errors, as that's an API consideration.

I think probably the right thing to do here is prevent all the content on the page from showing. So instead of doing a ternary here, wrap the whole `return` statement in an if condition, and if the `user` prop is empty, just display a message saying they'll need to log in (you could even `<Link to="/login">`). Ideally, we'd keep the header and footer regardless.

Datguyovrder: I pushed my recent code. I got rid of the 401 part of the message. Right now it only displays when a user is logged in. Eslint is having issues that I have no idea how to fix. It fails the circleCI tests but passes through Docker.

vhoof: Yeah, I think it has to do with using the huge ternary and the way it's formatted. I think we'll want to solve that a different way, probably using an actual `if` statement and two returns. Ternaries are meant for small things, otherwise they become pretty unreadable.

You could break the content out into a function, like
```
renderContent() {
  if (user) {
    return (
       // all the content here
    );
  }
  return (
    <h1>You need to log in to view this page.</h1>
  );
}
```

And then just called `renderContent` in `render`'s `return`.

nickharriscodes: Thanks for the feedback. Most recent commit should be a fix/refactor for this.

PR #320, opened 2019-08-14
Adds conditional renderButtons function

Datguyovrder:

When a user is logged in, the user is able to see the edit and delete buttons on the recipes that the user has personally added (picture 1). The edit and delete buttons will not appear on recipes that do not belong to the user (picture 2)

screen_shot_2019-08-13_at_7 05 02_pm

screen_shot_2019-08-13_at_6 51 17_pm

### Checklist For Submitter


frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -69,16 +88,9 @@ class Recipe extends Component {
...
-            <img src="https://img.icons8.com/windows/32/ffa616/edit.png" alt="edit" />
-            <p id="edit-icon-text">Edit</p>
-          </button>
-          <button className="btn navbar-btn btn-lg cancel-button" type="button">
-            <img src="https://img.icons8.com/windows/32/ffa616/cancel.png" alt="delete" />
-            <p id="delete-icon-text">Delete</p>
-          </button>
-        </div>
+        <span>
+          { user ? renderButtons() : 'null' }
```
vhoof: I don't think you need the `span` tag. I think you also want `null` without the single quotes here, since you don't want to return the string "null".

frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -69,16 +88,7 @@ class Recipe extends Component {
...
-          <button className="btn navbar-btn btn-lg edit-button" type="button">
-            <img src="https://img.icons8.com/windows/32/ffa616/edit.png" alt="edit" />
-            <p id="edit-icon-text">Edit</p>
-          </button>
-          <button className="btn navbar-btn btn-lg cancel-button" type="button">
-            <img src="https://img.icons8.com/windows/32/ffa616/cancel.png" alt="delete" />
-            <p id="delete-icon-text">Delete</p>
-          </button>
-        </div>
+        { user ? renderButtons() : null }
```
H12: I think having a ternary here, and an if statement in the function is a little redundant. You can probably consolidate things by doing `user && user.id === userId` here and removing the if statement from your `renderButtons` function.

frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -34,16 +34,32 @@ class Recipe extends Component {
...
   }
 
   render() {
     const { user, setJwt } = this.props;
     const {
-      directions, title, ingredients, snippet, difficulty, duration, servings, photo,
+      directions, title, ingredients, snippet, difficulty, duration, servings, photo, userId,
     } = this.state;
 
+    const renderButtons = function () {
```
vhoof: I think keeping this as an arrow function the way you had it before is preferable, especially to maintain consistency with all of the other functions. The linter was complaining because the function is just pure JSX, so you skip the return call and just wrap the body in parentheses. So it can just be:

```
const renderButtons = () => (
  // JSX here
);
```

Alternatively, since it's not doing anything dynamic, you could just convert it from a function to a JSX string by removing the `() =>` bit, and just use `renderButtons` without the parentheses on line 88.

frontend/package-lock.json
```diff
@@ -1099,7 +1099,7 @@
       "requires": {
```
vhoof: Looks like some package-lock changes still made it in. Maybe we should reset these?

vhoof: I also think a number of these changes made it in separately, so maybe reset these changes and then pull from staging?

PR #321, opened 2019-08-15
Only get user info when there's a JWT

vhoof:

When I first wrote the login code in App.js, I thought I'd be clever and save a couple lines of code. Instead of failing early when the JWT is null, I let the API request go through and just set the user to `null` when it inevitably failed.

Unfortunately, React logs the 401 it gets in this case, which is confusing when you're trying to develop/debug things. It's also potentially annoying for a more sophisticated end user.

This change prevents us from trying to get the current user's data when there is no logged in user.

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -20,9 +20,11 @@ const App = () => {
   useEffect(() => {
     if (jwt == null) {
       sessionStorage.removeItem('jwt');
-    } else {
-      sessionStorage.setItem('jwt', jwt);
+      setUser(null);
+      return;
     }
+
+    sessionStorage.setItem('jwt', jwt);
     getCurrentUser()
       .then(response => setUser(response.data))
       .catch(() => setUser(null));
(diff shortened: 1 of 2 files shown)


vhoof:
One other thing I noted is that we're presently using `sessionStorage`, which means you get logged out if you close the site and reopen it a moment later in a new browser tab. Wondering if we should consider switching to `localStorage`.

frontend/src/App.js
```diff
@@ -20,9 +20,11 @@ const App = () => {
   useEffect(() => {
     if (jwt == null) {
```
H12: Now that we have lodash, I think we should be a little more explicit here. Double-equals type coercion hurts my brain.

frontend/src/App.js
```diff
@@ -20,9 +20,11 @@ const App = () => {
...
       sessionStorage.removeItem('jwt');
-    } else {
-      sessionStorage.setItem('jwt', jwt);
+      setUser(null);
+      return;
     }
+
+    sessionStorage.setItem('jwt', jwt);
```
H12: Yep, I agree, we should almost certainly be using `localStorage`.

PR #323, opened 2019-08-16
password is being captured on the frontend

andreaMT15:
### Checklist For Submitter


frontend/src/containers/ProfilePage/ProfilePage.js
```diff
@@ -90,7 +90,10 @@ class ProfilePage extends Component {
   }
 
   render() {
+    // console.log(this.props.match);
+
```
andreaMT15: I started working on trying to get the dropdown to link to the settings tab by using URL parameters. I was following along with something I saw online. But, was struggling and figured I would table it until later. That is why the line above is commented out.

frontend/src/lib/apiClient.js
```diff
@@ -76,6 +76,12 @@ const resetPassword = (resetPasswordToken, password) => apiClient.put('/api/v1/p
   },
 });
 
+const resetLoginPassword = password => apiClient.put('/api/v1/password', {
```
vhoof: I'd be inclined to name this something like `resetCurrentUserPassword` and potentially rename the other function to `resetPasswordByToken`.

PR #325, opened 2019-08-18
Fix a bunch of React warnings

vhoof:

We've been fairly consistently getting a bunch of react warnings in the console throughout the site.  I caused some of these with children-cloning shenanigans in `AuthCard`, which I planned to fix. We were also somewhat inconsistent about how we used `required`, and I fixed that. 

Along the way, I also upgraded react-router to ensure it doesn't spew warnings due to using deprecated lifecycle methods. In the same vein, I changed the `RecipeSearchList` to a functional component to help avoid the use of deprecated methods.

There were also a few places where we used `map` without assigning a key, which are now fixed. And I fixed the lodash-related linter warnings for the files I touched.


frontend/.eslintrc
```diff
@@ -12,7 +12,7 @@
     "jsx-a11y/label-has-for": 0,
     "jsx-a11y/no-autofocus": 0,
     "jsx-a11y/no-static-element-interactions": 0,
-    "lodash/prefer-lodash-method": 1,
+    "lodash/prefer-lodash-method": [1, {"ignoreObjects": ["React\\.Children"]}],
```
vhoof: `React.Children` isn't really an array but rather an interface to a more complicated data structure. So when the linter complains about replacing `React.Children.map` and its ilk with lodash, it's mistaken.

Issue #293, opened 2019-06-11
To Fix: Styling on page indicator on the Admin dashboard


tkhutch:
Currently in app:

![image.png](https://images.zenhubusercontent.com/5c2ea74fa3ed691d821afa37/5c619507-dc70-4f44-ae67-e82e3090b12c)

Desired Styling:

![image.png](https://images.zenhubusercontent.com/5c2ea74fa3ed691d821afa37/f2ca3815-908d-4c1f-9f0e-78ff449f2043)

So that the number doesn't look so odd, let's keep the < and > on either sides of the list of pages but make them not hyperlinks when there aren't actually more results

sbudisav:
I got the pagination arrows working pretty easy. They are colored (blue) when there are more pages available to click, and black and un-clickable when there are no more pages. 

Functionality is fine but the styling is still boxed. I'm going to go hunting around for a styling that matches the picture shown in the ticket. 
<img width="209" alt="Screen Shot 2019-07-24 at 7 22 31 PM" src="https://user-images.githubusercontent.com/32229605/61837112-e981ff80-ae48-11e9-80f0-a3a4ecff0a14.png">

PR #350, opened 2019-09-16
Feature/293

rim23ma:
### Ticket Number:
#293

Changed style for pagination by overriding some properties. Now it is always < and > icons and  they are not hyperlinks when there aren't actually more results.

Screen Shot 2019-09-16 at 1 12 04 AM

Screen Shot 2019-09-16 at 1 04 16 AM

### Checklist For Submitter

### DIFF
--- a/frontend/src/components/AdminControls/Recipes.js
+++ b/frontend/src/components/AdminControls/Recipes.js
@@ -111,10 +111,12 @@ const Recipes = ({ recipes }) => {
 
   const options = {
     hideSizePerPage: true,
+    alwaysShowAllBtns: true,
+    withFirstAndLast: false,
     sizePerPageList: [
       {
         text: '5',
-        value: 20,
+        value: 5,
       },
       {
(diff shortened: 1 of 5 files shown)


vhoof:
It looks like the merge commit `cdcf7a1` is causing some issues. Can you do something like:
```
git checkout staging
git pull
git checkout feature/293
git rebase --onto staging HEAD~4
git push -f origin HEAD
```

That should, I think, fix your commit history.

frontend/src/components/AdminControls/Recipes.js
```diff
@@ -111,6 +111,8 @@ const Recipes = ({ recipes }) => {
 
   const options = {
     hideSizePerPage: true,
+    alwaysShowAllBtns: true,
+    withFirstAndLast: false,
```
vhoof: Discussed this in person, but think these settings also need to be applied to the user table, as well.

PR #352, opened 2019-09-16
Feature/333/new recipe2

andreaMT15:
### Checklist For Submitter

### DIFF
--- a/app/controllers/api/v1/recipes_controller.rb
+++ b/app/controllers/api/v1/recipes_controller.rb
@@ -1,4 +1,5 @@
 class Api::V1::RecipesController < ApplicationController
+  before_action :authenticate_user!, :except => [:show, :index]
 
   def show
     @recipe = Recipe.find_by!(id: params[:id])
@@ -16,15 +17,13 @@ class Api::V1::RecipesController < ApplicationController
 
   def create
     @recipe = Recipe.new(recipe_params)
-    @recipe.save
-    @recipe
-
+    @recipe.user = current_user
+    @recipe.save  
+    render json: @recipe
   end
 
   def recipe_params
-   params.require(:recipe).permit(:title, :snippet, :difficulty, :duration, :servings, :user_id )
+   params.require(:recipe).permit(:title, :snippet, :difficulty, :duration, :servings )
   end
 
 end
-
-
--- a/app/controllers/api/v1/recipes_controller.rb
+++ b/app/controllers/api/v1/recipes_controller.rb
@@ -13,4 +13,18 @@ class Api::V1::RecipesController < ApplicationController
     end
       render json: @recipes, :include => [:user]
   end
+
+  def create
+    @recipe = Recipe.new(recipe_params)
+    @recipe.save
+    @recipe
+
+  end
+
+  def recipe_params
+   params.require(:recipe).permit(:title, :snippet, :difficulty, :duration, :servings, :user_id )
+  end
+
 end
+
+
--- a/frontend/src/components/EditRecipe/EditRecipeForm.js
+++ b/frontend/src/components/EditRecipe/EditRecipeForm.js
@@ -1,86 +1,85 @@
-import React, { Component } from 'react';
-
+import React, { useState } from 'react';
+import { createRecipe } from '../../lib/apiClient';
 import Footer from '../Footer/Footer';
 import './EditRecipeForm.css';
 
-class EditRecipeForm extends Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      title: '', // eslint-disable-line react/no-unused-state
-      snippet: '', // eslint-disable-line react/no-unused-state
-    };
-    this.handleInputChange = this.handleInputChange.bind(this);
-  }
+const EditRecipeForm = ({ user }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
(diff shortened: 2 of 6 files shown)


frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -1,86 +1,85 @@
...
+  const [servings, setServings] = useState('2');
 
-  handleInputChange(event) {
-    this.setState({
-      [event.target.name]: event.target.value,
+  const handleSubmit = () => {
+    const currentUserId = user.id;
+    createRecipe({
+      title, description, difficulty, duration, servings, currentUserId,
     });
```
nickharriscodes: I think this method should give the user some sort of confirmation when their recipe has been submitted. Maybe the create recipe boxes disappear and are replaced with a message saying "Recipe Created!"

app/controllers/api/v1/recipes_controller.rb
```diff
@@ -13,4 +13,18 @@ def index
     end
       render json: @recipes, :include => [:user]
   end
+
+  def create
```
vhoof: Since only users can create recipes, we'll want to reflect that in the API by only allowing authenticated requests to this endpoint. Check out https://github.com/plataformatec/devise/wiki/How-To:-Define-resource-actions-that-require-authentication-using-routes.rb

Essentially, we can either require authentication at the route level or at the controller level. I think controller-level might be more succinct, but either way is fine.

app/controllers/api/v1/recipes_controller.rb
```diff
@@ -13,4 +13,18 @@ def index
...
+    @recipe = Recipe.new(recipe_params)
+    @recipe.save
```
vhoof: Once we're authenticated, you could call `@recipe.user = current_user`, as `current_user` will be set as part of the authentication stuff. You'll want to do that before you save, and this will be preferable to/more secure than allowing the client to provide the user_id.

app/controllers/api/v1/recipes_controller.rb
```diff
@@ -13,4 +13,18 @@ def index
...
+    @recipe
+
+  end
+
+  def recipe_params
+   params.require(:recipe).permit(:title, :snippet, :difficulty, :duration, :servings, :user_id )
```
vhoof: If we allow `user_id` as a param here, a malicious user could create recipes under any user's ID. Under the circumstances, that's probably not a big deal, but best security practices should prevent us from allowing it.

app/controllers/api/v1/recipes_controller.rb
```diff
@@ -13,4 +13,18 @@ def index
...
+  end
+
 end
+
+
```
vhoof: Got some extra blank lines here.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -1,86 +1,85 @@
...
-    this.state = {
-      title: '', // eslint-disable-line react/no-unused-state
-      snippet: '', // eslint-disable-line react/no-unused-state
-    };
-    this.handleInputChange = this.handleInputChange.bind(this);
-  }
+const EditRecipeForm = ({ user }) => {
+  const [title, setTitle] = useState('');
+  const [description, setDescription] = useState('');
+  const [difficulty, setDifficulty] = useState('Advanced');
```
vhoof: I think we have been using all caps for difficulty-related stuff.

frontend/src/lib/apiClient.js
```diff
@@ -75,6 +75,18 @@ const getRecipes = (search) => {
 
 const getRecipe = id => apiClient.get(`api/v1/recipes/${id}`);
 
+const createRecipe = recipe => apiClient.post('api/v1/recipes#create', {
```
vhoof: The `create` part shouldn't be in the URL. `api/v1/recipes` points to the `RecipesController`, and `create` in Rails is just the method that gets called when you `POST`.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -1,86 +1,85 @@
...
+  const [duration, setDuration] = useState('60 mins');
```
vhoof: I believe duration is an integer on the backend, so this will need to be converted before being submitted.

frontend/src/lib/apiClient.js
```diff
@@ -75,6 +75,18 @@ const getRecipes = (search) => {
...
+  recipe: {
+    title: recipe.title,
+    snippet: recipe.description,
+    difficulty: recipe.difficulty,
+    duration: recipe.duration,
+    servings: recipe.servings,
+    user_id: recipe.currentUserId,
```
vhoof: As mentioned in the backend code, you'll want to remove this. You will need to send the authorization header, though.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -1,86 +1,85 @@
...
+  const [servings, setServings] = useState('2');
```
vhoof: Servings is also an integer, so you may need to convert that before submitting it.

frontend/src/components/EditRecipe/EditRecipeForm.js
vhoof: Yeah, I think ideally we actually push the user to the created recipe page.

PR #359, opened 2019-10-09
Feature/346

andreaMT15:
Added a comments section to each recipe. When logged in, users can leave comments. When not logged in, users can see the comments, but won't be able to write one. 

### Checklist For Submitter


andreaMT15:
@vhoof  We're having a small issue after the comment gets posted. It tries to load the recipe again, but then it doesn't render any of the recipe information. Or it just crashes. We're not getting any errors in the logs. Could you take a look to see if anything stands out to you?

app/controllers/api/v1/comments_controller.rb
```diff
@@ -0,0 +1,26 @@
+class Api::V1::CommentsController < ApplicationController
+    before_action :authenticate_user!, :except => [:show, :index]
+    
+    def index
+        @comments = Comment.all
```
vhoof: Since the comments are tied to a particular recipe, you probably don't want *all* the comments, just the ones that match `params[:recipe_id]`.

app/controllers/api/v1/comments_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+        render json: @comments
+    end
+
+    def show
```
vhoof: This probably isn't needed, since there probably isn't a use case for viewing a single comment.

config/routes.rb
```diff
@@ -17,8 +17,11 @@
 
   namespace :api do
     namespace :v1 do
-      resources :recipes
+      resources :recipes do
+        resources :comments
+      end
       resources :users
+      resources :comments
```
vhoof: I see you have this in here twice. I think it made sense for testing, but we probably just want the one nested under recipes?

frontend/src/components/RecipeComments/RecipeComments.js
```diff
@@ -0,0 +1,67 @@
+import React, { useState } from 'react';
+import map from 'lodash/map';
+import { createComment } from '../../lib/apiClient';
+import './RecipeComments.css';
+
+const RecipeComments = ({ comments, recipeId, user }) => {
+  const [comment, setComment] = useState('');
+  const handleChange = (event) => {
```
vhoof: I'd put a blank line above this for readability.

frontend/src/components/RecipeComments/RecipeComments.js
```diff
@@ -0,0 +1,67 @@
...
+    event.preventDefault();
+    setComment(event.target.value);
+  };
+
+  const handleSubmit = () => {
+    createComment(comment, recipeId);
```
vhoof: You might need to pass `event` to the function and call `event.preventDefault();` above this to prevent some strange behavior.

frontend/src/components/RecipeComments/RecipeComments.js
```diff
@@ -0,0 +1,67 @@
...
+      commentArr.push(comments[i].body);
+      commentsArr.push(commentArr);
+    }
+    return (
+      <div>
+        {map(commentsArr, value => (
+          <div className="comment-content">
+            <hr />
+            <p><strong>{value[0]}</strong></p>
+            <p>{value[1]}</p>
```
vhoof: Instead of doing the loop above, why not just map over `comments` and use `${value.user.first_name} ${value.user.last_name}` and `value.body` here?

frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -160,6 +166,7 @@ class Recipe extends Component {
             }
           </ul>
         </div>
+        { comments ? <RecipeComments comments={comments} recipeId={recipeId} user={user} /> : null }
```
vhoof: I might use `!isNull(comments)` (with `isNull` coming from lodash). Right now, this has the right behavior, but the intent is a little murky (especially given that the truthiness of `[]` varies from language to language).

frontend/src/lib/apiClient.js
```diff
@@ -102,6 +102,19 @@ const createRecipe = recipe => apiClient.post('api/v1/recipes', {
   },
 });
 
+// COMMENTS
+
+const getComments = id => apiClient.get(`api/v1/comments/${id}`);
```
vhoof: I think you probably want this to take a `recipeID` and call `api/v1/recipes/${recipeID}/comments`. This presently gets a single comment.

frontend/src/components/RecipeComments/RecipeComments.js
nickharriscodes: Originally, I had tried to map over the comments array directly, but I got some errors because the .map method from vanilla JS didn't like that there were objects nested inside the array. So I used the for loop to extract the necessary data and put it into arrays inside an array.

app/controllers/api/v1/comments_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Api::V1::CommentsController < ApplicationController
+    before_action :authenticate_user!, :except => [:show, :index]
+    
+    def index
+        @comments = Comment.find(params[:recipe_id])
```
vhoof: I think this would find comments where `comments.id = :recipe_id`, which is probably not what you want. I think probably `Comment.where(recipe_id: params[:recipe_id]).all` is what you want.

frontend/src/components/RecipeComments/RecipeComments.js
vhoof: Hmm, I'm not sure what that would be. I think mapping over comments directly would be preferred here, and maybe there was some other bug causing problems with map.

frontend/src/lib/apiClient.js
```diff
@@ -102,6 +102,19 @@ const createRecipe = recipe => apiClient.post('api/v1/recipes', {
   },
 });
 
+// COMMENTS
+
+const getComments = recipeId => apiClient.get(`api/v1/recipes/${recipeId}`);
```
vhoof: You'll still want `/comments` at the end of this. And you may not need this method at all, now that I look deeper, since you're returning the comments with the recipe.

frontend/src/containers/RecipePage/Recipe.js
nickharriscodes: Would we need to create a separated if statement in order to employ isNull, or can we do it within the ternary?

app/controllers/api/v1/comments_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      @comment.recipe_id = params[:recipe_id]
+      @comment.save
+      render json: @comment
+    end
+
+    def comment_params
+      params.require(:comment).permit(:body)
+    end
+
+  end
```
H12: NitPick: Plz add a newline at the end of this file.

test/models/comment_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class CommentTest < ActiveSupport::TestCase
+  # test "the truth" do
```
H12: 💀

frontend/src/components/RecipeComments/RecipeComments.js
H12: Agreed. If you want more context, the accepted answer to this stackoverflow should be helpful:
https://stackoverflow.com/questions/28479239/setting-onsubmit-in-react-js

(Same deal as with your `handleChange` function).

PR #362, opened 2019-10-17
Feature/360/comment fix

andreaMT15:
### Ticket Number:
360

Comments are reloaded without the page refreshing. Also, added moment.js package in order to format timestamp. 

### Checklist For Submitter

Screen Shot 2019-10-16 at 7 10 06 PM

### DIFF
--- a/frontend/src/components/RecipeComments/RecipeComments.js
+++ b/frontend/src/components/RecipeComments/RecipeComments.js
@@ -37,7 +37,7 @@ const RecipeComments = ({
   const renderComments = () => (
     <div>
       {map(comments, value => (
-        <div className="comment-content">
+        <div key={value.id} className="comment-content">
           <hr />
           <p>
             <strong>
--- a/frontend/src/components/RecipeComments/RecipeComments.js
+++ b/frontend/src/components/RecipeComments/RecipeComments.js
@@ -12,10 +12,10 @@ const RecipeComments = ({
     const months = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October',	'November', 'December'];
     const d = new Date(date);
     let hours = d.getHours();
-    var amOrPm = hours >= 12 ? 'PM' : 'AM';
+    const amOrPm = hours >= 12 ? 'PM' : 'AM';
     hours = (hours % 12) || 12;
-    return months[d.getMonth()] + ' ' + d.getDate() + ', ' + d.getFullYear() + ' ' + hours + ':' + d.getMinutes() + ' ' + amOrPm;
-  }
+    return (`${months[d.getMonth()]} ${d.getDate()} ${d.getFullYear()} ${hours}:${d.getMinutes()} ${amOrPm}`);
+  };
 
   const handleChange = (event) => {
     event.preventDefault();
@@ -24,7 +24,6 @@ const RecipeComments = ({
 
   const submitComment = async () => {
     setComment('');
-    console.log(comments[1].created_at);
     await createComment(comment, recipeId);
     reloadComments();
   };
--- a/frontend/src/components/RecipeComments/RecipeComments.js
+++ b/frontend/src/components/RecipeComments/RecipeComments.js
@@ -1,6 +1,5 @@
 import React, { useState } from 'react';
 import map from 'lodash/map';
-import * as moment from 'moment';
 import { createComment } from '../../lib/apiClient';
 import './RecipeComments.css';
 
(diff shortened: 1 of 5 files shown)


andreaMT15:
@vhoof I accidentally pushed my new branch onto this branch. I'll look into doing a rebase to fix this tomorrow.

frontend/src/components/RecipeComments/RecipeComments.js
```diff
@@ -1,38 +1,52 @@
 import React, { useState } from 'react';
 import map from 'lodash/map';
+import * as moment from 'moment';
```
vhoof: It may seem crazy to worry about it, but I wonder if it'd be better to use [date-fns](https://date-fns.org/) instead of moment. Moment is.. pretty bulky.

frontend/src/components/RecipeComments/RecipeComments.js
```diff
@@ -1,38 +1,52 @@
...
   };
 
-  const handleSubmit = () => {
-    createComment(comment, recipeId);
+  const submitComment = async () => {
+    setComment('');
+    await createComment(comment, recipeId);
+    const newData = { body: comment };
+    const newCommentData = { ...comments, ...newData };
+    reRenderComments(newCommentData);
```
vhoof: What's the `newData` and `newCommentData` stuff for?

I might also call `reRenderComments` something like `reloadComments` to be more in line with what it's doing. The re-render is just a side effect of changing the state.

frontend/src/components/RecipeComments/RecipeComments.js
```diff
@@ -1,38 +1,52 @@
...
   const handleChange = (event) => {
     event.preventDefault();
     setComment(event.target.value);
   };
 
-  const handleSubmit = () => {
-    createComment(comment, recipeId);
+  const submitComment = async () => {
+    setComment('');
+    await createComment(comment, recipeId);
```
vhoof: Didn't we just set `comment` to an empty string?

nickharriscodes: setComment('') resets the comment form/text area. Without that bit, the form does not reset.

frontend/src/components/RecipeComments/RecipeComments.js
nickharriscodes: Yep, those variables were unnecessary... I forget what I was doing there. Took them out, and renamed that function.

frontend/src/components/RecipeComments/RecipeComments.js
nickharriscodes: Will check out and try to implement date-fns!

frontend/src/components/RecipeComments/RecipeComments.js
vhoof: Right. And I know this works due to the asynchronous nature of `setComment`. It's just somewhat counterintuitive to set the `comment` to an empty string, then rely on its previous value on the next line, if that makes sense.

nickharriscodes: I think it would still work if I move the setComment down to after reloadComments();

Would that address/fix this?

nickharriscodes: I went ahead and moved that setComment down below reloadComments. Does that sufficiently improve the logic/flow of this?

frontend/src/components/RecipeComments/RecipeComments.js
```diff
@@ -3,36 +3,66 @@ import map from 'lodash/map';
...
 import './RecipeComments.css';
 
-const RecipeComments = ({ comments, recipeId, user }) => {
+const RecipeComments = ({
+  comments, recipeId, user, reloadComments,
+}) => {
   const [comment, setComment] = useState('');
 
+  const formatMinutes = (minute) => {
+    const zero = 0;
```
H12: This seems unnecessary. You should just be about to do something like this on line 14:
```
return `0${minute}`;
```

frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -56,6 +55,15 @@ class Recipe extends Component {
       userId, comments,
     } = this.state;
 
+    const reloadComments = () => {
+      const { match: { params } } = this.props;
+      getRecipe(params.id).then((response) => {
+        this.setState({
+          comments: response.data.comments.reverse(),
+        });
```
H12: I think this can go on one line.

frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -56,6 +55,15 @@ class Recipe extends Component {
       userId, comments,
     } = this.state;
 
+    const reloadComments = () => {
+      const { match: { params } } = this.props;
+      getRecipe(params.id).then((response) => {
+        this.setState({
+          comments: response.data.comments.reverse(),
```
H12: It's generally best-practice to do this kind of sorting on the back-end.

PR #367, opened 2019-10-24
Feature/343/tags search

rim23ma:
1. Added search bar for tags
2. Tags are coming from database
3. User can delete the chosen tag from create recipe page but not from created recipe. 

### Checklist For Submitter

* [X] I have screenshotted new UI changes and attached them to this PR
Screen Shot 2019-10-23 at 8 11 49 PM


frontend/src/components/TagsBar/SingleTag/SingleTag.js
```diff
@@ -1,9 +1,14 @@
...
 
-const SingleTag = ({ tag }) => (
-  <div className="single-tag">
-    {tag}
+const SingleTag = ({ tag, deleteSelTag, showDeleteButton }) => (
+  <div>
+    <div className="single-tag">
+      <div className="tag-name">{tag}</div>
+      {showDeleteButton
+        ? <div className="delete-tag" onClick={() => deleteSelTag(tag)}>x</div> : <div className="hide" />
```
vhoof: Since this is `display: none`, you're effectively inserting a completely invisible div (aka nothing) here. You could just put `null` in the else condition, which will actually insert nothing.

app/controllers/api/v1/recipes_controller.rb
```diff
@@ -16,14 +16,21 @@ def index
   end
 
   def create
-    @recipe = Recipe.new(recipe_params)
+    rp = recipe_params
+    tags = rp.delete(:tags)
+
+    @recipe = Recipe.new(rp)
+    tags.each do |tag|
+      @recipe.tags << Tag.find(tag["id"])
```
vhoof: I think `tag[:id]` would be preferable here. This is a Ruby thing.

app/controllers/api/v1/tags_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def index
+    if params[:recipe_id]
+      @tags = Recipe.find_by!(id: params[:recipe_id]).tags
+    else
+      @tags = Tag.all
+    end
+      render json: @tags
+   end
+
```
vhoof: I'd remove this empty line.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -1,25 +1,72 @@
...
   const [servings, setServings] = useState('2');
+  const [tags, setTags] = useState([]);
+  const [selectedTags, setSelectedTags] = useState([]);
+  const [selectedTagsWithId, setSelectedTagsWithId] = useState([]);
+
+  useEffect(() => {
+    getTags().then((data) => {
+      if (tags.length === 0) setTags(data);
+    });
+  }, [tags]);
```
vhoof: I think this will cause this to get called twice (or in an infinite loop if there are no tags in the database). This last line means that we run this function whenever `tags` changes, and we set `tags` inside the function. So it causes itself to be called again. It's not the worst thing in the world, but it's not great either.

frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -20,13 +20,15 @@ class Recipe extends Component {
...
       photo: '',
+      tags: [],
     };
   }
 
 
   componentDidMount() {
     window.scrollTo(0, 0);
     const { match: { params } } = this.props;
+    const { match: { params: { id } } } = this.props;
```
vhoof: It looks like we're doing this twice, in both cases to access `id`. We should probably just do one or the other (either extract `id` or `params.id`), and use it wherever we need it.

frontend/src/containers/RecipePage/Recipe.js
```diff
@@ -41,12 +43,17 @@ class Recipe extends Component {
         userId: response.data.user_id,
       });
     });
+    getRecipeTags(id).then((data) => {
```
vhoof: I think we should be able to modify the recipe `show` method in Rails to include the tags, so we don't need to make this second call at all.

frontend/src/components/EditRecipe/EditRecipeForm.js
vhoof: We could use a `refreshTags` state object that starts as `true`, and then check that inside the `useEffect` here, and set it to false when we refresh the tags. Then use `refreshTags` as the catalyst here on line 39. Like:

```
  useEffect(() => {
    if (refreshTags) {
      getTags().then((data) => {
        if (tags.length === 0) setTags(data);
      });
      setRefreshTags(false);
    }
  }, [refreshTags]);
```

Or something like that.

vhoof:
Looks like there are now quite a few conflicts with staging due to merging the comments functionality. :/ We'll need to fix those before we can merge.

vhoof:
One final issue. When you select a tag, it causes some weird behavior in the search bar:

Screen Shot 2019-10-28 at 6 17 21 PM
Screen Shot 2019-10-28 at 6 17 33 PM

If you click the X in the search bar, it deletes all of the tags you've added up to that point. I assume that's related to the react-select stuff, but maybe that's something we can control.

frontend/src/lib/apiClient.js
```diff
@@ -95,13 +95,22 @@ const createRecipe = recipe => apiClient.post('api/v1/recipes', {
...
     Authorization: localStorage.jwt,
   },
 });
 
+// TAGS
+
+const getTags = () => apiClient.get('/api/v1/tags')
+  .then(results => results.data);
+
+const getRecipeTags = id => apiClient.get(`/api/v1/recipes/${id}/tags`)
```
vhoof: I think we may need to delete this function now that we're not using it (at least for now). It's causing CircleCI to fail due to linting errors:
```
./src/containers/RecipePage/Recipe.js
  Line 9:  'getRecipeTags' is defined but never used  no-unused-vars
```

frontend/src/components/EditRecipe/EditRecipeForm.css
```diff
@@ -22,6 +22,21 @@
   border-color: #9b9b9b;
 }
 
+select {
```
vhoof: I know these `select`, `input`, and `label` things here are related to react-select, but can we provide custom classes on those bits (like `.create-recipe-tags-select`, etc)? As is, we're styling every `select`, `input`, and `label` on the site, which I don't think is what we want.

db/seeds.rb
```diff
@@ -141,6 +141,30 @@
...
+Tag.create(
+ title: "Snack"
+)
+
+tag1 = Tag.find(1)
+tag2 = Tag.find(2)
+
+recipe1 = Recipe.find(1)
+recipe1.tags << tag1
+recipe1.tags << tag2
```
vhoof: We could actually clean this up a bit. The `create` method returns the thing it creates, so you can say `breakfast = Tag.create(title: 'Breakfast')`. We could also do that with one of the recipes, like `recipe = Recipe.create(...)`.

Then instead of adding random tags to a random recipe, we could say `recipe.tags = [breakfast, lunch]`. That way we know which recipe has been given which tags, and it's easier to test.

frontend/src/components/TagsBar/TagsBar.js
```diff
@@ -1,26 +1,23 @@
...
+  <div className="tags-bar col-sm-10 offset-sm-1">
+    {map(tags, tag => (
+      <SingleTag
+        tag={tag.title}
+        key={tag.id}
+        deleteSelTag={deleteSelTag}
+        showDeleteButton={showDeleteButton}
+      />
+    ))}
+    <AddTag />
```
nickharriscodes: Shouldn't the` <AddTag />` component be hidden when you're just viewing the recipe? It seems like it would only be necessary when on the create recipe form.

rim23ma: We didn't work with Add Tag component. We have it on next ticket 344, where we hid it.

frontend/src/components/EditRecipe/EditRecipeForm.css
Nicoleta13: I deleted the styling for " select, input and label " because they were styled already in the classes. So these were just double styling.

Screen Shot 2019-10-29 at 10 26 05 PM

db/seeds.rb
Nicoleta13: Screen Shot 2019-10-30 at 5 26 32 PM

app/controllers/api/v1/tags_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  def index
+    if params[:recipe_id]
+      @tags = Recipe.find_by!(id: params[:recipe_id]).tags
+    else
+      @tags = Tag.all
+    end
+    render json: @tags
+  end
+end
```
vhoof: Can you go into your VSCode settings and enable final newlines? From VS Code, press `Command-,`, then click on `Text Editor`, then `Files`. Scroll down a bit and check the box under `Insert Final Newline`.

It's good practice to have a newline at the end of every file, as it's a standard, and certain editors and parsers can choke on the lack of them. To fix that, we'll need to add a blank line at the end of this file, but with that VSCode setting, it'll do it automatically when you save the file.

app/models/recipe.rb
```diff
@@ -13,5 +13,6 @@ class Recipe < ApplicationRecord
     }
   )
   belongs_to :user
+  has_and_belongs_to_many :tags
   has_many :comments, -> { order('created_at DESC') }
-end
+end
```
vhoof: Need a newline at the end here, too.

db/seeds.rb
```diff
@@ -151,4 +169,4 @@
   body: "This is another test comment",
   user_id: 2,
   recipe_id: 20
-)
+)
```
vhoof: Need a newline at the end here, too.

frontend/src/components/EditRecipe/EditRecipeForm.css
```diff
@@ -43,11 +43,18 @@
...
+
 .editRecipeForm i.fas.fa-utensil-spoon {
   padding-right: 15px;
   padding-bottom: 0px;
 }
 
+.hide {
+  display: none;
+}
 .recipe-submit-container{
```
vhoof: Can we put a blank line above this?

frontend/src/components/EditRecipe/EditRecipeForm.css
```diff
@@ -62,10 +69,20 @@
...
+
 .form-dropdown{
   display: flex;
   flex-direction: column;
   align-items: center;
   justify-content: space-between;
   margin-bottom: 50px;
 }
+
+textarea {
```
vhoof: We should use a class here, as I don't think we want to style every `textarea` on the site.

frontend/src/components/EditRecipe/EditRecipeForm.css
```diff
@@ -62,10 +69,20 @@
...
+  resize: none;
+  border-color: #9b9b9b;
+}
```
vhoof: Need a newline at the end here, too.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -1,26 +1,80 @@
...
 const EditRecipeForm = ({ history }) => {
   const [title, setTitle] = useState('');
   const [description, setDescription] = useState('');
   const [difficulty, setDifficulty] = useState('ADVANCED');
   const [duration, setDuration] = useState('30 mins');
   const [servings, setServings] = useState('2');
+  const [tags, setTags] = useState([]);
+  const [selectedTags, setSelectedTags] = useState([]);
+  const [selectedTagsWithId, setSelectedTagsWithId] = useState([]);
+  const [refreshTags, setRefreshTags] = useState([]);
```
vhoof: I think you want this to start off as `true`.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -36,6 +90,36 @@ const EditRecipeForm = ({ history }) => {
...
           </label>
         </div>
+        <div className="search-tag col-6 offset-3">
+          <label className="detail-labels" htmlFor="tags">
+            Tags
+            <Select
+              className="st-search-input"
+              styles={customStyles}
+              value={selectedTags}
+              options={tags.map((tag) => {
```
vhoof: Can we use lodash for this `map` call to avoid linter warnings?

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -36,6 +90,36 @@ const EditRecipeForm = ({ history }) => {
...
+                setSelectedTags(selected);
+              }
+            }
+              placeholder="Search and tag"
+              isMulti
+            />
+            <TagsBar
+              deleteSelTag={handleDelete}
+              tags={selectedTagsWithId}
+              showDeleteButton="yes"
```
vhoof: I think you want `true` here, rather than `"yes"`. To JavaScript, any string is truthy, so even if you put `"no"` here, it'd have the same effect.

frontend/src/components/TagsBar/SingleTag/SingleTag.js
```diff
@@ -1,9 +1,14 @@
...
 
-const SingleTag = ({ tag }) => (
-  <div className="single-tag">
-    {tag}
+const SingleTag = ({ tag, deleteSelTag, showDeleteButton }) => (
+  <div>
+    <div className="single-tag">
+      <div className="tag-name">{tag}</div>
+      {showDeleteButton
+        ? <div className="delete-tag" onClick={() => deleteSelTag(tag)}>x</div> : null
```
vhoof: Can we move the `: null` part to the next line to aid readability?

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -36,6 +90,36 @@ const EditRecipeForm = ({ history }) => {
...
+              })}
+              onChange={(selected) => {
+                setSelectedTags(selected);
+              }
+            }
+              placeholder="Search and tag"
+              isMulti
+            />
+            <TagsBar
+              deleteSelTag={handleDelete}
```
vhoof: What is `SelTag`?

PR #372, opened 2019-10-31
Feature/344/add tags

rim23ma:
1. User can add a new tag in database
2. It doesn't duplicate tags
3. Created tag automatically becomes selected by user
4. User can add tag only on Create Recipe page

### Checklist For Submitter


app/controllers/api/v1/tags_controller.rb
```diff
@@ -6,6 +6,15 @@ def index
...
+  def create
+    @tag = Tag.new(tag_params)
+    @tag.save
+    render json: @tag
   end
-end
\ No newline at end of file
+end
+
+def tag_params
```
H12: This should be a private method with in the tags controller.

app/controllers/api/v1/tags_controller.rb
```diff
@@ -6,6 +6,15 @@ def index
     else
       @tags = Tag.all
     end
-    render json: @tags
+      render json: @tags
+   end
```
H12: The spacing here is kinda messy. I think we want something formatted closer to this:
```ruby
class controller
  def action
    if condition
      # code
    else
      # more code
    end

    # render
  end

  def other_action
    #code
  end
end
```

PR #375, opened 2019-11-01
BottomSignUp component is not rendering if user signed in

rim23ma:
1. BottomSignUp component is not rendering if user signed in.

### Checklist For Submitter


frontend/src/containers/Homepage/Homepage.js
```diff
@@ -40,9 +40,12 @@ const Homepage = ({ location, user, setJwt }) => {
           <div className="row">
             <RecipeSearchList searchTerm={searchTerm} />
           </div>
-          <div className="row">
-            <BottomSignUp />
-          </div>
+          { user ? null
```
vhoof: Can we use `isNil(user)` from [lodash](https://lodash.com/docs/#isNil) here? That'll help clarify the intent and make it less error-prone.

PR #382, opened 2019-11-05
Feature/328/merged staging remove sign up

rim23ma:
1. SignUpBottom component is not shown for logged-in users

### Checklist For Submitter

### DIFF
--- a/frontend/src/containers/Homepage/Homepage.js
+++ b/frontend/src/containers/Homepage/Homepage.js
@@ -19,8 +19,6 @@ const Homepage = ({ location, user, setJwt }) => {
     scrollToRef(myRef);
   };
 
-  console.log('isNil', isNil(user));
-
   return (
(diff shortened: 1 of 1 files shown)


frontend/src/containers/Homepage/Homepage.js
```diff
@@ -40,9 +41,13 @@ const Homepage = ({ location, user, setJwt }) => {
...
           </div>
-          <div className="row">
-            <BottomSignUp />
-          </div>
+          { isNil(user) ? (
+            <div className="row">
+              <BottomSignUp />
+            </div>
+          )
+            : null
```
H12: The spacing/formatting is a little off here. I think something like this would be a little cleaner:
```
{ isNil(user) ? (
    <div className="row">
      <BottomSignUp />
    </div>
  ) : null
}

PR #383, opened 2019-11-07
Feature/341

rim23ma:
1. Only admin has an access to AdminPanel
2. If user doesn't have admin status or non logged in than it will redirect to homepage.

### Checklist For Submitter

ezgif com-gif-maker

### DIFF
--- a/frontend/src/containers/AdminPanel/AdminPanel.js
+++ b/frontend/src/containers/AdminPanel/AdminPanel.js
@@ -40,9 +40,13 @@ class AdminPanel extends Component {
       });
     });
 
-    setTimeout(() => {
-      this.setState({ render: true });
-    }, 2000);
+    const { user } = this.props;
+
+    if (isNil(user)) {
+      setTimeout(() => {
+        this.setState({ render: true });
+      }, 2000);
+    } else this.setState({ render: true });
   }
 
   refreshUsers = () => {
@@ -79,7 +83,9 @@ class AdminPanel extends Component {
     } = this.state;
     if (render) {
       if (isNil(user)) return (<Redirect to="/login" />);
+
       if (!user.is_admin) return (<Redirect to="/" />);
+
       if (!isNil(user)) {
         return (
           <div className="admin-panel-container">
--- a/frontend/src/containers/AdminPanel/AdminPanel.js
+++ b/frontend/src/containers/AdminPanel/AdminPanel.js
@@ -3,7 +3,9 @@ import React, { Component } from 'react';
 import {
   TabContent, TabPane, Nav, NavItem, NavLink, Button,
 } from 'reactstrap';
+import isNil from 'lodash/isNil';
 import { Helmet } from 'react-helmet';
+import { Redirect } from 'react-router-dom';
 import Header from '../../components/Header/Header';
 import Footer from '../../components/Footer/Footer';
 import Recipes from '../../components/AdminControls/Recipes';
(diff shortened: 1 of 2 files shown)


frontend/src/containers/AdminPanel/AdminPanel.js
```diff
@@ -68,6 +70,11 @@ class AdminPanel extends Component {
   render() {
     const { user, setJwt } = this.props;
     const { activeTab, recipes, users } = this.state;
+    const isAdmin = isNil(user) ? 'loadingPage' : user.is_admin;
```
H12: I don't think `isAdmin` should be responsible for determining whether the pace is loading, since that's a pretty different concept.

frontend/src/containers/AdminPanel/AdminPanel.js
```diff
@@ -68,6 +70,11 @@ class AdminPanel extends Component {
...
+
+    if (!localStorage.jwt) return <Redirect to="/" />;
+    if (isAdmin === 'loadingPage') return null;
+    if (!isAdmin) return <Redirect to="/" />;
```
H12: I think you can consolidate this with the condition on line 75, since you're redirecting to `/` in both cases.

vhoof:
I think since the problem is that it takes some time to get the user from the API, maybe just introducing a delay before the redirect? So like if user is null, wait a second or two, then redirect to the login page. If the user isn't null, either redirect to the home page or display the admin panel, depending on their is_admin status.

frontend/src/containers/AdminPanel/AdminPanel.js
```diff
@@ -67,60 +74,71 @@ class AdminPanel extends Component {
...
     const { user, setJwt } = this.props;
-    const { activeTab, recipes, users } = this.state;
+    const {
+      activeTab, recipes, users, render,
+    } = this.state;
+    let renderContainer = false;
+    if (render) {
+      if (isNil(user)) renderContainer = <Redirect to="/" />;
+      if (!isNil(user)) {
+        renderContainer = !user.is_admin ? <Redirect to="/" />
```
vhoof: Maybe instead of this `renderContainer` thing, we can just directly `return` whatever we want to return? Something like:
```
if (render) {
  if (isNil(user)) {
    return (<Redirect to="/login" />);
  }

  if (!user.is_admin) {
    return (<Redirect to="/" />);
  }

  return (
    // long thing here
  )
}

return null;
```

frontend/src/containers/AdminPanel/AdminPanel.js
```diff
@@ -36,6 +39,10 @@ class AdminPanel extends Component {
         users: this.createFullName(data),
       });
     });
+
+    setTimeout(() => {
+      this.setState({ render: true });
+    }, 2000);
```
vhoof: If we already have the user object, I'd just set `render` to true automatically. I'd only wait if you don't have the user.

frontend/src/containers/AdminPanel/AdminPanel.js
```diff
@@ -36,6 +39,14 @@ class AdminPanel extends Component {
         users: this.createFullName(data),
       });
     });
+
+    const { user } = this.props;
+
+    if (isNil(user)) {
+      setTimeout(() => {
```
H12: This feels kinda hacky, but I'm unsure of a better way to fix it. I've tried some things locally to no avail, so I guess it's fine for now.

@vhoof we may want to backlog a ticket for making this less hacky at some point.

vhoof: Yeah, I'm sorting it out in a different branch.

PR #389, opened 2019-11-08
Feature/373/edit recipes 2

nickharriscodes:
Allow users to edit recipes that they have created. CreateRecipe components dynamically render with selected recipe's parameters, user can adjust and submit. 

### Checklist For Submitter


vhoof:
You don't want to hit the API more than you have to. You can request the recipe once at the top (in `CreateRecipe`, in this case) then pass it down to the other components. You can then seed all your state objects in the form from the recipe you already have (like pass that instead of `id`, then `const [title, setTitle] = useState(recipe.title);`, etc.)

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
```diff
@@ -4,11 +4,13 @@ import reject from 'lodash/reject';
...
 import TagsBar from '../TagsBar/TagsBar';
-import { createRecipe, getTags } from '../../lib/apiClient';
+import {
+  createRecipe, editRecipe, getTags,
+} from '../../lib/apiClient';
 import barChart from '../../assets/bar-chart.png';
 import Footer from '../Footer/Footer';
-import './EditRecipeForm.css';
-import EditDirections from './EditDirections/EditDirections';
+import './CreateRecipeForm.css';
```
vhoof: I'm just curious, any reason the names of everything were changed to `CreateRecipe...`? It feels weird in a PR that changes them to add editing capabilities. :)

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
```diff
@@ -42,7 +44,32 @@ const EditRecipeForm = ({ history }) => {
   const [selectedTagsWithId, setSelectedTagsWithId] = useState([]);
   const [refreshTags, setRefreshTags] = useState(true);
   const [ingredients, setIngredients] = useState(['']);
+  const [isEditing, setIsEditing] = useState(false);
 
+  const handleTitleChange = (x) => {
+    setTitle(x);
+    setIsEditing(true);
+  };
```
vhoof: Can we add some blank lines between all these function definitions, here and in the ingredient/direction files?

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
```diff
@@ -42,7 +44,32 @@ const EditRecipeForm = ({ history }) => {
...
+  useEffect(() => {
+    if (currentRecipe && !isEditing) {
+      setTitle(currentRecipe.title);
+      setDescription(currentRecipe.snippet);
+      setDifficulty(currentRecipe.difficulty);
+      setSelectedTagsWithId(currentRecipe.tags);
+      setDuration(`${currentRecipe.duration} minutes`);
+      setServings(currentRecipe.servings);
+      // console.log(currentRecipe);
```
vhoof: Can we remove the comment here?

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
```diff
@@ -42,7 +44,32 @@ const EditRecipeForm = ({ history }) => {
...
   const [refreshTags, setRefreshTags] = useState(true);
   const [ingredients, setIngredients] = useState(['']);
+  const [isEditing, setIsEditing] = useState(false);
 
+  const handleTitleChange = (x) => {
+    setTitle(x);
+    setIsEditing(true);
+  };
+  const handleDescriptionChange = (x) => {
+    // console.log('checkvalue ', x, description);
```
vhoof: Can we remove the comment here?

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
```diff
@@ -42,7 +44,32 @@ const EditRecipeForm = ({ history }) => {
   const [selectedTagsWithId, setSelectedTagsWithId] = useState([]);
   const [refreshTags, setRefreshTags] = useState(true);
   const [ingredients, setIngredients] = useState(['']);
+  const [isEditing, setIsEditing] = useState(false);
```
vhoof: What is the purpose of `isEditing`? If we're trying to prevent the `useEffect`s from overwriting changes, we can make those hooks [fire conditionally](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect).

frontend/src/components/Header/Header.js
```diff
@@ -7,8 +7,12 @@ import HeaderDropdown from '../HeaderDropdown/HeaderDropdown';
 import './Header.css';
 
 const Header = ({
-  history, user, setJwt, showSearchBar, scrollToRecipes,
+  history, user, setJwt, showSearchBar, match, scrollToRecipes,
 }) => {
+  const refreshCreatePage = () => {
+    history.push('/recipe/new');
+    window.location.reload(false);
+  };
```
vhoof: Can we put a blank line between all the function definitions on this page?

frontend/src/components/Header/Header.js
```diff
@@ -17,6 +21,13 @@ const Header = ({
     }
   };
   const loginOrCreateButton = () => {
+    if (user && match.params.id) {
```
vhoof: This feels a little hacky, as there might be other pages that use ids in the URL. I'm also curious if there's a more elegant way to handle this problem than forcing a reload, but not sure off the top of my head what might be quick to implement. But I don't necessarily want to block stuff getting merged at this point.

frontend/src/containers/CreateRecipe/CreateRecipe.js
```diff
@@ -1,24 +1,45 @@
...
+      <div className="row justify-content-md-center">
+        <h1 className="text-center">
+          {heading}
+        </h1>
+      </div>
+      <div className="row">
+        <div className="edit-recipe-form">
+          <CreateRecipeForm
+            user={user}
+            currentRecipe={match.params.id >= 0 ? currentRecipe : null}
```
vhoof: I think you could just send `currentRecipe` and set that to null to begin with on line 9.

frontend/src/containers/CreateRecipe/CreateRecipe.js
```diff
@@ -1,24 +1,45 @@
...
-const CreateRecipe = ({ user, setJwt }) => (
-  <div className="createRecipePage container-fluid">
-    <div className="row">
-      <div className="createRecipePage-header">
-        <Header showSearchBar user={user} setJwt={setJwt} />
+const CreateRecipe = ({ user, setJwt, match }) => {
+  const [heading, setHeading] = useState('Create Recipe');
+  const [currentRecipe, setCurrentRecipe] = useState();
+  const [isEditing, setIsEditing] = useState(false);
+  useEffect(() => {
```
vhoof: Can we put some blank lines around this `useEffect` call?

frontend/src/components/CreateRecipeForm/CreateRecipeForm.css
```diff
@@ -1,9 +1,10 @@
 .recipe-details {
-  border-color: #9b9b9b;
+  border-color: rgb(204, 204, 204);
```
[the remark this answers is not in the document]
nickharriscodes: I was using the Digital Color Meter on my app to see what color l needed to set the border to, and it shows the color in rgb. I'll change to #CCCCCC.

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
nickharriscodes: I changed them when I first started this ticket thinking that I'd need to create a separate component for editing recipes, and the form used to create recipes was called EditRecipeForm.

frontend/src/components/Header/Header.js
nickharriscodes: From the user's perspective, I don't know that it would feel unnatural or unexpected for the page to reload if they're going from editing a recipe to creating an entirely new one. I feel like almost any other solution would require many more lines of code so idk....

Also I think the only other pages that have ids in the url are the individual recipe pages, and the CreateRecipeForm component would have to load up anyway when navigating there anyway, right? Will have to look into this more tomorrow.

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
nickharriscodes: @vhoof The way the ingredients and directions are set up, without isEditing, the respective handleChange functions won't allow the state to change/won't allow any editing in the text inputs.

vhoof: Yeah, I think that's because the useEffect hooks fire every time the component gets updated (which it will every time you change `directions`). If you wrote those like:

```
useEffect(() => {
  // do some stuff
}, [currentRecipe]);
```

This would cause that hook to only fire when the `currentRecipe` object changes, which it shouldn't unless you change the URL. Unless there's something else going on I'm not understanding?

vhoof:
Other than the `isEditing` stuff, it all looks good to me. @H12 want to take a look? Maybe there's something there I'm missing. Other than that, I'm happy to approve and merge it.

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
H12: @nickharriscodes Did you look into this suggestion? I think conditional firing of the hooks would be the preferred way forward here.

H12:
@vhoof I agree with you on the `isEditing` thing. It looks like your suggestion would work.

frontend/src/components/CreateRecipeForm/CreateRecipeForm.js
nickharriscodes: @H12 ran into some errors... trying again right now

H12:
@vhoof After talking w/ @nickharriscodes in person, it looks like there was a problem with the conditional useEffect hook.

I think leaving the isEditing thing in is fine.

vhoof:
Sounds good. We have some merge conflicts. If those aren't sorted out in the next couple days, I'll figure them out and get this merged over the weekend.

PR #407, opened 2019-11-15
Feature/378/non-submitable

andreaMT15:
### Ticket Number:
378

User will not be allowed to submit a recipe unless they enter a title, description, 1 ingredient, and 1 set of directions. 

Screen Shot 2019-11-14 at 8 43 21 PM

### Checklist For Submitter

### DIFF
--- a/frontend/src/components/EditRecipe/EditRecipeForm.css
+++ b/frontend/src/components/EditRecipe/EditRecipeForm.css
@@ -95,3 +95,8 @@
 .recipe-description {
   resize: none;
 }
+
+.errorMsg{
+  color: red;
+  font-size: .7em;
+}
--- a/frontend/src/components/EditRecipe/EditRecipeForm.css
+++ b/frontend/src/components/EditRecipe/EditRecipeForm.css
@@ -96,7 +96,7 @@
   resize: none;
 }
 
-.errorMsg{
+.error-message{
   color: red;
   font-size: .7em;
 }
--- a/frontend/src/components/EditRecipe/EditRecipeForm.js
+++ b/frontend/src/components/EditRecipe/EditRecipeForm.js
@@ -65,9 +65,7 @@ const EditRecipeForm = ({ history }) => {
     setSelectedTagsWithId(tagsWithIdObject);
   }, [selectedTags]);
 
-  const handleSubmit = () => {
-    const parsedDuration = parseInt(duration, 10);
-    const parsedServings = parseInt(servings, 10);
+  const validate = () => {
     if (!title) {
       setTitleError('Title cannot be blank');
     } if (!description) {
@@ -76,6 +74,14 @@ const EditRecipeForm = ({ history }) => {
       setIngredientError('Ingredients cannot be blank');
     } if (directions.length <= 1) {
       setDirectionsError('Directions cannot be blank');
+    }
+  };
+
+  const handleSubmit = () => {
+    const parsedDuration = parseInt(duration, 10);
+    const parsedServings = parseInt(servings, 10);
+    if (!title || !description || ingredients.length <= 1 || directions.length <= 1) {
+      validate();
     } else {
       createRecipe({
         title,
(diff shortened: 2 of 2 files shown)


frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -100,14 +114,16 @@ const EditRecipeForm = ({ history }) => {
...
           </label>
+          <div className="errorMsg">{title ? '' : titleError}</div>
         </div>
         <div className="form-snippet col-6 offset-3">
           <label className="detail-labels" htmlFor="snippet">
             Recipe Description
-            <textarea className="form-control recipe-details recipe-description" id="description" name="description" value={description} rows="4" onChange={e => setDescription(e.target.value)} />
+            <textarea className="form-control recipe-details recipe-description" id="description" name="description" value={description} rows="4" onChange={e => setDescription(e.target.value)} required />
           </label>
+          <div className="errorMsg">{description ? '' : descriptionError}</div>
```
H12: I think you can just do `{descriptionError}`, because the `descriptionError ` is an empty string by default.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -100,14 +114,16 @@ const EditRecipeForm = ({ history }) => {
         <div className="form-title col-4 offset-4">
           <label className="detail-labels" htmlFor="title">
             Recipe Title
-            <input className="form-control input-sm recipe-details" id="title" type="text" name="title" value={title} onChange={e => setTitle(e.target.value)} />
+            <input className="form-control input-sm recipe-details" id="title" type="text" name="title" value={title} onChange={e => setTitle(e.target.value)} required />
           </label>
+          <div className="errorMsg">{title ? '' : titleError}</div>
```
H12: I think you can just do `{titleError}`, because the `titleError` is an empty string by default.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -183,10 +199,12 @@ const EditRecipeForm = ({ history }) => {
         ingredients={ingredients}
         setIngredients={setIngredients}
       />
+      <div className="errorMsg">{ingredients.length <= 1 ? ingredientsError : ''}</div>
```
H12: I think you can just do `{ingredientsError}`, because the `ingredientsError ` is an empty string by default.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -183,10 +199,12 @@ const EditRecipeForm = ({ history }) => {
...
       <EditDirections
         directions={directions}
         setDirections={setDirections}
       />
+      <div className="errorMsg">{directions.length <= 1 ? directionsError : ''}</div>
```
H12: I think you can just do `{directionsError}`, because the `directionsError ` is an empty string by default.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -63,17 +68,26 @@ const EditRecipeForm = ({ history }) => {
...
-      tags: selectedTagsWithId,
-      directions,
-      ingredients,
-    }).then(response => history.push(`/recipe/${response.data.id}`));
+    if (!title) {
+      setTitleError('Title cannot be blank');
+    } if (!description) {
+      setDescriptionError('Description cannot be blank');
+    } if (ingredients.length <= 1) {
+      setIngredientError('ingredients cannot be blank');
```
H12: You're calling `setIngredientError` here, but it's defined as `setIngredientsError` above.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -63,17 +68,26 @@ const EditRecipeForm = ({ history }) => {
...
+    } if (directions.length <= 1) {
+      setDirectionsError('directions cannot be blank');
+    } else {
+      createRecipe({
```
H12: I think we might need something more here. As is, it looks like the only thing stopping `createRecipe` from getting called is `directions.length` being greater than one.

frontend/src/components/EditRecipe/EditRecipeForm.css
```diff
@@ -95,3 +95,8 @@
 .recipe-description {
   resize: none;
 }
+
+.errorMsg{
```
H12: Generally, it's best practice to use `kebab-case` for CSS classes.

frontend/src/components/EditRecipe/EditRecipeForm.js
```diff
@@ -60,20 +65,35 @@ const EditRecipeForm = ({ history }) => {
...
-      description,
-      difficulty,
-      duration: parsedDuration,
-      servings: parsedServings,
-      tags: selectedTagsWithId,
-      directions,
-      ingredients,
-    }).then(response => history.push(`/recipe/${response.data.id}`));
+    if (!title || !description || ingredients.length <= 1 || directions.length <= 1) {
+      validate();
```
vhoof: Personally, I'd be inclined to have `validate` return true or false based on whether an error was generated or not. Then `if (validate()) { createRecipe...` would be cleaner, and we wouldn't be checking the same if conditions twice.

But if @H12 is cool with the changes, I'm down to not block it.

H12: I agree with that feedback.

andreaMT15:
@H12 can you please re-review?

PR #411, opened 2019-11-17
Add node-sass

H12:
Trying to add node-sass so we can use scss. Running into docker problems with OS versions. Node-sass is getting built with MacOS (apparently) which prevents it from running within the container.

### Checklist For Submitter


vhoof:
Do we need node-sass installed in the root, or can we move it up into the frontend folder?

H12:
@vhoof It should be fine to have it in the frontend folder. The fact that it's in the root was a mistake on my part.

vhoof:
Should be good to go now. If you run `npm install` in your frontend folder, that should give you the mac version locally. If you run `docker-compose run -w /app/frontend web npm install`, that should give you the linux version in the container.

PR #413, opened 2019-11-19
Feature/390/link to admin

nickharriscodes:
Added links to allow admins to access the admin panel. 

### Checklist For Submitter

* [ x] I have documented all new functionality
* [ x] I have screenshotted new UI changes and attached them to this PR
Screen Shot 2019-11-18 at 8 07 48 PM
Screen Shot 2019-11-18 at 8 08 30 PM


frontend/src/components/HeaderDropdown/HeaderDropdown.js
```diff
@@ -40,6 +46,7 @@ const HeaderDropdown = ({ history, user, setJwt }) => {
         <Link className="header-dropdown-item" to="/profile#settings">
           <DropdownItem className="header-dropdown-item">Settings</DropdownItem>
         </Link>
+        {user.is_admin === true ? showAdminLink() : null}
```
vhoof: You could just use `user.is_admin` here. It's already boolean, so the `=== true` is redundant.

Same on the other page.

PR #414, opened 2019-11-21
Feature/345

rim23ma:
1.  Display the recipes that belong to the current user

### Checklist For Submitter

Screen Shot 2019-11-20 at 7 55 49 PM


app/controllers/api/v1/recipes_controller.rb
```diff
@@ -1,36 +1,41 @@
 class Api::V1::RecipesController < ApplicationController
   before_action :authenticate_user!, :except => [:show, :index]
 
-  def show
-    @recipe = Recipe.find_by!(id: params[:id])
-    render json: @recipe, :include => [{comments: {include: :user }}, :tags]
-  end
+    def show
```
vhoof: I assume this was Prettier or some other automated thing, but it looks like you've changed the indentation throughout this file. It was correct before, so if we can revert those changes, that'd be awesome.

app/controllers/api/v1/recipes_controller.rb
```diff
@@ -1,36 +1,41 @@
...
-  def index
-    if params[:search]
-      @recipes = Recipe.search(params[:search])
-    else
-      @recipes = Recipe.includes(:user).all
+    def index
+      if params[:search]
+        @recipes = Recipe.search(params[:search])
+      else
+        if params[:user_id]
```
vhoof: I would try to avoid nesting if statements unless there's a good reason. In this case, you could just use [`elsif`](https://www.tutorialspoint.com/ruby/ruby_if_else.htm).

frontend/src/containers/ProfilePage/ProfilePage.js
```diff
@@ -11,71 +11,18 @@ import Footer from '../../components/Footer/Footer';
...
-      user_id: 1,
-    },
-  ];
+  const [recipes, setRecipes] = useState([]);
 
-  useEffect(() => {
+  useEffect(async () => {
     window.scrollTo(0, 0);
+    const recipesData = await getUserRecipes(user.id);
+    setRecipes(recipesData);
```
vhoof: It seems like using the promise structure would be better here, like:
```
getUserRecipes(user.id).then((data) => setRecipes(data));
```

PR #415, opened 2019-11-22
Feature/393/admin comment delete

nickharriscodes:
When logged in as an admin, user has the option of deleting comments. UI shows an "X" button in bottom corner of each recipe comment which brings up a modal asking if user would like to delete comment. 

### Checklist For Submitter

Screen Shot 2019-11-21 at 6 29 07 PM


frontend/src/components/RecipeComments/RecipeComments.js
```diff
@@ -2,11 +2,18 @@ import React, { useState } from 'react';
 import map from 'lodash/map';
 import { createComment } from '../../lib/apiClient';
 import './RecipeComments.css';
+import DeleteCommentModal from './DeleteCommentModal/DeleteCommentModal';
 
 const RecipeComments = ({
   comments, recipeId, user, reloadComments,
 }) => {
   const [comment, setComment] = useState('');
+  const [deleteModal, setDeleteModal] = useState(false);
```
vhoof: Nitpick: Maybe calling this `showDeleteModal` would be clearer.

app/controllers/api/v1/comments_controller.rb
```diff
@@ -14,6 +14,11 @@ def create
       render json: @comment
     end
 
+    def destroy
```
vhoof: Something we haven't been doing so far is making sure only admins can do certain things at the API level, too. Might be good to do that here. You could add a `return head(:forbidden) unless current_user.is_admin` at the top of this function, which will cause this to return a 403 if they're not an admin.

PR #433, opened 2019-12-19
Feature/395/limit recipes

rim23ma:
1. Only 20 first recipes are displayed.
2. Button "load more" loads 20 more recipes.
3. Css for button "load more".

### Checklist For Submitter

Screen Shot 2019-12-18 at 8 26 24 PM

### DIFF
--- a/frontend/src/components/RecipeSearchList/RecipeSearchList.js
+++ b/frontend/src/components/RecipeSearchList/RecipeSearchList.js
@@ -7,10 +7,29 @@ import { getRecipes } from '../../lib/apiClient';
 
 const RecipeSearchList = ({ searchTerm }) => {
   const [recipes, setRecipes] = useState([]);
+  const [page, setPage] = useState(1);
+  const [newSearchTerm, setnewSearchTerm] = useState(false);
+
   useEffect(() => {
-    getRecipes(searchTerm)
+    setnewSearchTerm(true);
+    setPage(1);
+    getRecipes(1, searchTerm)
       .then(data => setRecipes(data));
   }, [searchTerm]);
+
+  useEffect(() => {
+    if (!newSearchTerm) {
+      getRecipes(page, searchTerm)
+        .then(data => setRecipes(recipes.concat(data)));
+    }
+  }, [page]);
+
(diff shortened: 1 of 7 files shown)


frontend/src/components/RecipeSearchList/RecipeSearchList.js
```diff
@@ -4,27 +4,50 @@ import RecipeCard from '../RecipeCard/RecipeCard';
...
+    getRecipes(1, searchTerm)
       .then(data => setRecipes(data));
   }, [searchTerm]);
 
+  useEffect(() => {
+    if (!newSearchTerm) {
+      getRecipes(page, searchTerm)
+        .then(data => setRecipes(recipes.concat(data)));
+    }
+  }, [page]);
```
vhoof: What if we used `[page, searchTerm]` here? That would trigger a `getRecipes` call when `page` OR `searchTerm` changes. Then you could remove the `getRecipes` call in the other `useEffect`, the if call in this one, and the `newSearchTerm` variable entirely.

rim23ma: Hi Bryan! I am trying to change a code, but I have a question, what should I put in if statements when I get rid of the newSearchTerm? Thank you!

const [recipes, setRecipes] = useState([]);
const [page, setPage] = useState(1);

useEffect(() => {
setPage(1);
setRecipes([]);
}, [searchTerm]);

useEffect(() => {
getRecipes(page, searchTerm)
.then(data => {
if (?????) setRecipes(recipes.concat(data));
else if (??????) setRecipes(data);
});
}, [page, searchTerm]);

const loadMoreRecipes = () => {
setPage(page + 1);
};

vhoof: Ahh, hmm. I may have misread the code the first time. Let me take another look and get back to you.

vhoof: After thinking about this some more, let's just leave it as is. I don't want to go round the bend trying to DRY this up a tiny amount.

PR #452, opened 2020-01-24
Feature/399/edit delete recipe by admin

### LINKED ISSUE #399: Edit and Delete recipe buttons on admin dashboard should be functional
Edit: Load the Edit recipe page. Would be nice if in this circumstance we could redirect back to the admin dashboard after the changes are saved.

Delete: pop a delete modal:

Clicking delete deletes the recipe

rim23ma:
### Ticket Number
Closes #399

1. Admin can edit recipes from admin dashboard
2. Admin can delete recipes from admin dashboard

### Checklist For Submitter

ezgif com-video-to-gif

### DIFF
--- a/app/controllers/api/v1/recipes_controller.rb
+++ b/app/controllers/api/v1/recipes_controller.rb
@@ -47,8 +47,10 @@ class Api::V1::RecipesController < ApplicationController
 
   def destroy
     return head(:forbidden) unless current_user.is_admin
-    @recipe = Recipe.find_by!(id: params[:id])
-    @recipe.destroy
+    @recipe = Recipe.find(params[:id])
+    if @recipe
+      @recipe.destroy
+    end
   end
 
   def recipe_params
--- a/app/controllers/api/v1/recipes_controller.rb
+++ b/app/controllers/api/v1/recipes_controller.rb
@@ -40,9 +40,15 @@ class Api::V1::RecipesController < ApplicationController
     tags.each do |tag|
       db_tags << Tag.find(tag[:id])
     end
-   
+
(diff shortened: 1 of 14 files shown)


app/controllers/api/v1/recipes_controller.rb
```diff
@@ -40,11 +40,17 @@ def update
...
     end
-   
+
     @recipe.tags = db_tags
     @recipe.update(rp)
   end
 
+  def destroy
+    return head(:forbidden) unless current_user.is_admin
+    @recipe = Recipe.find_by!(id: params[:id])
```
H12: NitPick: I think it would be a little cleaner to just use `find`:
```ruby
Recipe.find(params[:id])
```

app/controllers/api/v1/recipes_controller.rb
```diff
@@ -40,11 +40,17 @@ def update
...
+    @recipe.destroy
```
vhoof: `find_by!` will cause this endpoint to generate an error if it doesn't find the recipe. But since we're trying to delete the recipe, we probably shouldn't error when it's not found. So I think you could use `Recipe.find(params[:id])`, and then `@recipe.destroy if @recipe`.

frontend/src/components/AdminControls/Recipes.js
```diff
@@ -33,6 +43,7 @@ const Recipes = ({ recipes }) => {
   );
 
   const recipeRow = ({
+    id,
```
vhoof: I think you need to pass this into `adminButtons` below on line 57. That's what CircleCI/the linter is choking on.

